PR #3, opened 2021-09-07
swap to ratio

ewilz:
Reviewer:

- [x] Re-orged oclif. If it looks good I can DRY. If not, lmk a better way to organize the CLI :)
- [x] `routeToAmountsRatio`: Calculate amountIn based on parameters. Could I be wielding sdk objects better?
- [x] typescript halp
- [x] @NoahZinsmeister  or @moodysalem  to check v3/math etc. (https://github.com/Uniswap/uniswap-smart-order-router/pull/3/files?file-filters%5B%5D=.ts#diff-2fe4a092f36b8091262f7035638dc00530816382f3a1cca86927a6fd712e9c04R116)

Design doc:
https://www.notion.so/uniswaplabs/Swap-Add-Routing-API-3b3ba4fa645f4bbcb389a7b2f94a5d45

==================
swap-to-ratio feature

- there is no swap-to-ratio feature

- there IS swap-to-ratio but with limited impl https://github.com/Uniswap/uniswap-smart-order-router/issues/1

- **Other information**:


cli/base-command.ts
```diff
@@ -253,95 +253,41 @@ export class UniswapSORCLI extends Command {
...
-        `Could not find route. ${
-          debug ? '' : 'Run in debug mode for more info'
-        }.`
-      );
-      return;
-    }
+      this.id == 'quote-to-ratio'
+        ? (this._swapToRatioRouter = router)
+        : (this._router = router);
+      }
```
willpote, line 262: Nit: It would be fine to just initialize both, not really any overhead or perf hit

cli/commands/quote-to-ratio.ts
```diff
@@ -0,0 +1,160 @@
...
+        : tokenAccessor.getTokenByAddress(token0Str)!;
+    const tokenOut: Currency =
+      token1Str == 'ETH'
+        ? Ether.onChain(chainId)
+        : tokenAccessor.getTokenByAddress(token1Str)!;
+
+    const tokenInBalance = parseAmount(token0BalanceStr, tokenIn);
+    const tokenOutBalance = parseAmount(token1BalanceStr, tokenOut);
+
+    const poolAccessor = await this.poolProvider?.getPools([
```
willpote, line 78: Shouldn't need the `?.` anymore now that you throw errors in the getters, here and below

src/routers/router.ts
```diff
@@ -56,3 +56,13 @@ export abstract class IRouter<RoutingConfig> {
     routingConfig?: RoutingConfig
   ): Promise<SwapRoute<TradeType.EXACT_OUTPUT> | null>;
 }
+
+export abstract class ISwapToRatio<RoutingConfig> {
```
willpote, line 60: Best if this is an interface, then AlphaRouter can implement it too (in addition to IRouter) so we'd have something like `class AlphaRouter implements IRouter, ISwapToRatio { }`

Also I think you forgot to update AlphaRouter so this is currently unused anyway

willpote (approved):
Couple nits but overall looks good

willpote (approved):

PR #6, opened 2021-09-20
Iterate if initial ratio does not match optimal

ewilz:
Possible to test functionality with CLI. 

Iterative Swap:
```
./bin/cli quote-to-ratio --token0 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 --token1 0xdac17f958d2ee523a2206206994597c13d831ec7 --feeAmount 3000 --recipient 0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B  --token0Balance 1000 --token1Balance 20000 --tickLower -120 --tickUpper 120
```

Good the first time:
```
./bin/cli quote-to-ratio --token0 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 --token1 0xdac17f958d2ee523a2206206994597c13d831ec7 --feeAmount 3000 --recipient 0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B  --token0Balance 1000 --token1Balance 2000 --tickLower -120 --tickUpper 120
```

**Todo**:
- [x] Limit the number of iterations to a maximum (is it possible to never land on something within error tolerance?)
- [x] Account for price changes to the target pool
- [ ] Assign position/pool object a blockNumber?

**Edgecases**
- [x] out of range positions
- [x] in range positions but swap moves price outside of position's range (above or below)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -147,23 +148,61 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+        console.log('outputBalance', outputBalance.toFixed(6))
+        console.log('amountToSwap', amountToSwap.toFixed(6))
+        console.log('inputAmount', swap.trade.inputAmount.toFixed(6))
+        console.log('outputAmount', swap.trade.outputAmount.toFixed(6))
+        console.log('inputBalanceUpdated', inputBalanceUpdated.toFixed(6))
+        console.log('outputBalanceUpdated',  outputBalanceUpdated.toFixed(6))
+        console.log('optimalRatio', optimalRatio.toFixed(6))
+        console.log('newRatio    ', newRatio.asFraction.toFixed(6))
+
+        ratioAchieved = newRatio.asFraction.divide(optimalRatio).subtract(1).lessThan(errorTolerance)
```
ewilz, line 201: Should clean this up. I'm checking that `newRatio` is within `errorTolerance %` of `optimalRatio`. I'm subtracting 1 (instead of `1 - newRatio/optimalRatio`) because newRatio is probably larger than optimalRatio since amountOut is probably less than we expected. 

However I don't want to always assume that, so I want the absolute value of this subtraction. Can I get the absolute value of a `Fraction` or potentially add that to its api?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -542,4 +564,38 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+
+    let optimalRatio =  new Fraction(
+      token0Proportion,
+      token1Proportion
+    )
+    if (!zeroForOne) optimalRatio = optimalRatio.invert()
+    return optimalRatio
+  }
+
+  private absoluteValue(fraction: Fraction): Fraction {
```
ewilz, line 595: could something like this live in the `Fraction` sdk module??

src/routers/alpha-router/alpha-router.ts
```diff
@@ -129,41 +130,62 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+              pool.token0 == position.pool.token0 &&
+              pool.token1 == position.pool.token1 &&
+              pool.fee == position.pool.fee
+            ) {
+              optimalRatio = this.calculateOptimalRatio(position, JSBI.BigInt(route.sqrtPriceX96AfterList[i]!.toString()))
+            }
+          })
+        })
+
+        ratioAchieved = this.absoluteValue(newRatio.asFraction.divide(optimalRatio).subtract(1)).lessThan(errorTolerance)
```
ewilz, line 184: can probably clean this line up...
checking that `newRatio` is within % `errorTolerance` of `optimalRatio`

src/routers/alpha-router/alpha-router.ts
```diff
@@ -129,41 +130,67 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
-        token1Balance,
-      )
+      const zeroForOne = new Fraction(token0Balance.quotient, token1Balance.quotient).greaterThan(optimalRatio)
+      if (!zeroForOne) optimalRatio = optimalRatio.invert()
+
+      const [inputBalance, outputBalance] = zeroForOne
+        ? [token0Balance, token1Balance]
+        : [token1Balance, token0Balance]
+
+      let exchangeRate: Fraction = zeroForOne ? position.pool.token0Price : position.pool.token1Price
```
willpote, line 145: token0Price and token1Price actually return a `Price` object, on which you get use `.quote(x)` to get an output amount (in terms of a CurrencyAmount object). I think this would simplify the `calculateRatioAmountIn` function below

src/routers/alpha-router/alpha-router.ts
```diff
@@ -129,41 +130,67 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+        )
+        if (!swap) {
+          throw new Error('did not fetch swap')
+        }
+
+        let inputBalanceUpdated = inputBalance.subtract(swap.trade.inputAmount)
+        let outputBalanceUpdated = outputBalance.add(swap.trade.outputAmount)
+        let newRatio = inputBalanceUpdated.divide(outputBalanceUpdated)
+
+        swap?.route.forEach(route => {
```
willpote, line 172: `?` not needed since we already checked it exists on L164

src/routers/alpha-router/alpha-router.ts
```diff
@@ -129,41 +130,67 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+                JSBI.BigInt(route.sqrtPriceX96AfterList[i]!.toString())
+              )
+            }
+          })
+        })
+
+        ratioAchieved =
+          newRatio.equalTo(optimalRatio) ||
+          this.absoluteValue(newRatio.asFraction.divide(optimalRatio).subtract(1)).lessThan(errorTolerance)
+        exchangeRate = swap.trade.outputAmount.divide(swap.trade.inputAmount)
```
willpote, line 190: Originally when we set exchangeRate we used the price of the pool we are adding liquidity for, but here we use the price implied by the swap (which may have been a multi-hop/split route).

Do we even need to update this at all? I thought exchange rate represents the starting price of the pool, so whenever we call `calculateRatioAmountIn` I expected it to be constant?

willpote (commented):
We need to make sure that whenever we get data on chain, or call `routeExactIn` everything is coming from the same block. 

You can specify block whenever you call something that goes onchain e.g. `poolProvider.getPool([...], { blockNumber: x })` and `this.routeExactIn`'s routingConfig lets you specify block

Not sure if you are still seeing weird results occasionally, but that may explain why. Infura will frequently return things from different blocks unless you specify it

src/routers/alpha-router/alpha-router.ts
ewilz: If put the same exchange rate back into the calculation, then we'll swap the same amount again. The iteration is mostly based on the fact that we did not get the amountOut we expected. So we try again at the exchange rate from the last trade, so that we have a better shot of calculating the correct amountOut.

Does that make sense?

src/routers/alpha-router/alpha-router.ts
ewilz: For some reason I'm having some trouble understanding this. I'd be calling quote on token0Price??

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +119,85 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+
+        let inputBalanceUpdated = inputBalance.subtract(swap.trade.inputAmount)
+        let outputBalanceUpdated = outputBalance.add(swap.trade.outputAmount)
+        let newRatio = inputBalanceUpdated.divide(outputBalanceUpdated)
+
+        swap.route.forEach(route => {
+          route.route.pools.forEach((pool, i) => {
+            if(
+              pool.token0 == position.pool.token0 &&
+              pool.token1 == position.pool.token1 &&
```
willpote, line 182: This won't work for comparing 2 objects, should use https://github.com/Uniswap/sdk-core/blob/main/src/entities/token.ts#L27

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +119,85 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+              pool.fee == position.pool.fee
+            ) {
+              optimalRatio = this.calculateOptimalRatio(
+                position,
+                JSBI.BigInt(route.sqrtPriceX96AfterList[i]!.toString())
+              )
```
willpote, line 188: Hmm won't this cause issues for our next iteration? 

Lets say on our first iteration our route uses the pool we are adding to and we update the optimal ratio. Then on our second iteration we'll have a new target optimal ratio, but we aren't guaranteed that the next `routeExactIn` will use the pool we are adding too (or even send the same % amount of the input to that pool) so our target ratio would be off

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +119,85 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+            }
+          })
+        })
+
+        ratioAchieved = (
+          newRatio.equalTo(optimalRatio) ||
+          this.absoluteValue(newRatio.asFraction.divide(optimalRatio).subtract(1)).lessThan(errorTolerance)
+        )
```
willpote, line 196: Do we have any sense of how many iterations it will take to converge? And if we are guaranteed to converge? 

Wondering 
1/ if it would be possible that we keep flipping between a ratio that is below our optimal and a ratio above our optimal causing an inf loop
2/ whether we should put a 'max iterations' to prevent infinite loops or the calculation taking too long / consuming too many resources
3/ whether we should terminate as soon as we are within the error threshold or not- continuing iterating could find us a swap that results in us being even closer to the optimal ratio and the user having less dust at the end

src/routers/alpha-router/alpha-router.ts
willpote: That makes sense, thanks

src/routers/alpha-router/alpha-router.ts
ewilz: When I do giant trades trading 200 million dollar values, I converge at about 4 trades. I think I still need to do some larger scoped tests against many different tokens/scenarios. 
1) investigating infinite loops I think is really important. Other than thinking hard through some potential scenarios, the best way I can think of is to try this with many different pools with many different amounts, but this is something I'm open to suggestions on.
2) this is in my todo, in the PR description unchecked. will add that,
3) I was going with the former, but open to ideas.

src/routers/alpha-router/alpha-router.ts
willpote: My thinking was that each iteration we essentially want to throw away the swap we just tried, reset to the state at the beginning of the block, and try a completely new swap.

If we use an updated sqrtPrice on the next iteration, we would be trying to find the ratio assuming we executed the swap from the previous iteration (which we wouldn't actually do).

Maybe I'm completely misunderstanding though, if so maybe we can hop on a call so you can set me right

src/routers/alpha-router/alpha-router.ts
willpote: 1. I'm not sure, but I think having max iterations will go a long way here
2. Ah ok great, sorry didnt see that
3. We could have something like a minimum iterations too. It could be tuned to give us a latency we're happy with. Feels like something we can follow up with though

willpote:
You'll have to pass the block number around so it will require changes to the code touched in this PR. I think its pretty important to address because getting routes from different blocks will break the algorithm if the price has moved, but don't mind if it comes in a separate PR

src/routers/alpha-router/alpha-router.ts
ewilz: It's currently going off the assumption that the next swap will swap a similar amount therefore having a similar effect. NOT that the last swap already happened, only that our next try will *also* move the price in the same way.

If we get to a point where our output quote is really lining up with what we expected it's only that we budged the sqrtPrice hence shifting optimalRatio, we need to try again swapping just a little bit more to reach that adjusted optimalRatio, but we need to know what that new optimalRatio is to do that, hence account for the new price. If the price is no longer pushed when trying to hit the new optimalRatio, then that's potentially infinite loop territory where we go back and forth.

willpote: I don't think we can assume that the next swap will go through the same pools though, so the sqrtPrice may not have moved at all. Changing the amount in a swap can result in very different route recommendations. This is the sort of example I'm worried about:

I have USDC. I want to add to USDC/USDT/5% pool

Iteration 1:

- `route.routeExactIn` recommends swap through the pool we are adding to (USDC —5%—>USDT )
- We use the new sqrtPrice of USDC/USDT/5% pool to compute optimalRatio. We use that to compare our ratio. We find it isn't close enough, so we proceed to next iteration.

Iteration 2:

- `route.routeExactIn` recommends a multi-hop through different pools (USDC —1%—>ETH—1%—>USDT )
- We use the optimalRatio from iteration 1 (which assumes we swapped on the USDC/USDT/5% pool, but we didn't) as our target ratio.
    - Therefore our target ratio is off?

willpote: I guess I'm wondering if it should be like:
```
let newOptimalRatio = optimalRatio
swap.route.forEach(route => {
          route.route.pools.forEach((pool, i) => {
            if(
              pool.token0.equals(position.pool.token0) &&
              pool.token1.equals(position.pool.token1) &&
              pool.fee == position.pool.fee
            ) {
              newOptimalRatio = this.calculateOptimalRatio(
                position,
                JSBI.BigInt(route.sqrtPriceX96AfterList[i]!.toString())
              )
            }
          })
        })

        ratioAchieved = (
          newRatio.equalTo(newOptimalRatio) ||
          this.absoluteValue(newRatio.asFraction.divide(newOptimalRatio).subtract(1)).lessThan(errorTolerance)
```

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +119,91 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
     position: Position,
+    errorTolerance: Fraction,
     swapConfig?: SwapConfig,
     routingConfig = DEFAULT_CONFIG
   ): Promise<SwapRoute<TradeType.EXACT_INPUT> | null> {
-      if (
-        token0Balance.currency.wrapped.address.toLowerCase()
-        > token1Balance.currency.wrapped.address.toLowerCase()
-      ) {
+      const maxIterations = 5
```
willpote, line 126: Can we make this parameterized? Will be good if we can play with this in the interface/routing-api without needing to do a release of the smart-order-router package.

Maybe a new SwapAndAddConfig object?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +119,91 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
+              )
+            }
+          })
+        })
+
+        ratioAchieved = this.absoluteValue(
+          newRatio.asFraction.divide(optimalRatio).subtract(1)
+        ).lessThan(errorTolerance)
+        exchangeRate = swap.trade.outputAmount.divide(swap.trade.inputAmount)
+      }
```
willpote, line 216: Would be great to log our newRatio, optimalRatio, and iteration number here

ewilz: added some logs, lmk if I should add anything, change the format!

src/routers/alpha-router/alpha-router.ts
ewilz: haha I'm not sure whether we are talking past each other or what! If we don't run the next iteration with newOptimalRatio, we'll never hit it, right? Because we keep aiming for optimalRatio all else being constant.

In your example above, I'd imagine there'd be an iteration3 that switches back to the previous `optimalRatio` and runs it again. The bad scenario here, is that the difference between optimalRatio and newOptimalRatio is what triggers the different trade path. My naive assumption is that getting the wrong exchange rate will mostly be determining that, esp in very liquid pools.

willpote: I accidently deleted my comment that had the code so just reposting it:

```
let newOptimalRatio = optimalRatio
swap.route.forEach(route => {
          route.route.pools.forEach((pool, i) => {
            if(
              pool.token0.equals(position.pool.token0) &&
              pool.token1.equals(position.pool.token1) &&
              pool.fee == position.pool.fee
            ) {
              newOptimalRatio = this.calculateOptimalRatio(
                position,
                JSBI.BigInt(route.sqrtPriceX96AfterList[i]!.toString())
              )
            }
          })
        })

        ratioAchieved = this.absoluteValue(
          newRatio.asFraction.divide(newOptimalRatio).subtract(1)
        ).lessThan(swapAndAddConfig.errorTolerance)
        exchangeRate = swap.trade.outputAmount.divide(swap.trade.inputAmount)
```

Well if we don't run the next iteration with `newOptimalRatio`, our exchange rate would still have changed, so I'd expect the next iteration to still get us closer to the optimal ratio. 

It would be the same situation as if our first iteration didn't use the pool we are adding to, so optimalRatio ends up being the same for our second iteration.

But yeah maybe we are talking past each other! Might be good to get @NoahZinsmeister thoughts on it too

src/routers/alpha-router/alpha-router.ts
ewilz: I think after further convo we decided against this?? Feel free to "unresolve" if I'm wrong here.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +124,114 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
-        true
-      )
-      const token1Proportion = SqrtPriceMath.getAmount1Delta(
-        sqrtPriceX96,
-        sqrtPriceX96Lower,
-        precision,
-        true
-      )
+      // exit early for out of range positions
+      if (position.pool.tickCurrent < position.tickLower) {
```
NoahZinsmeister, line 136: sorry a few questions here:

1. in this first case, where the current tick is below the lower, don't we want to be selling token1 for token0? and vice versa in the other case? as far as i can tell the code is reversed
2. assuming we sort out the above, i'm not sure we can early return. if we start out of range, i agree that we need to sell all of one asset for the other. but from working it out on paper, buying the asset we need will always push the market price closer to our desired range (assuming it executes in the pool we're targeting). so, simply selling all of our unwanted asset may push the price into our range, meaning our now 100/0 ratio is incorrect. am i missing something?

src/routers/alpha-router/functions/calculate-ratio-amount-in.ts
```diff
@@ -3,19 +3,19 @@ import { CurrencyAmount } from '../../../util/amounts';
...
 export function calculateRatioAmountIn(
   optimalRatio: Fraction,
-  token0Price: Fraction,
-  token0Balance: CurrencyAmount,
-  token1Balance: CurrencyAmount
+  inputTokenPrice: Fraction,
+  inputBalance: CurrencyAmount,
+  outputBalance: CurrencyAmount
 ): CurrencyAmount {
   // formula: amountToSwap = (token0Balance - (optimalRatio * token1Balance)) / ((optimalRatio * token0Price) + 1))
```
NoahZinsmeister, line 10: nit: should update this

src/routers/alpha-router/alpha-router.ts
```diff
@@ -119,51 +124,114 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
...
-        token0Balance,
-        token1Balance,
+      // set up parameters according to which token will be swapped
+      let zeroForOne = true
+      let optimalRatio = this.calculateOptimalRatio(
+        position,
+        position.pool.sqrtRatioX96,
+        zeroForOne
       )
+      zeroForOne = new Fraction(token0Balance.quotient, token1Balance.quotient).greaterThan(optimalRatio)
```
NoahZinsmeister, line 161: can put the let here

src/routers/alpha-router/alpha-router.ts
NoahZinsmeister: yeah definitely, though wouldn't consider it blocking

NoahZinsmeister (commented):
left some notes, will respond to the other comment thread as well

src/routers/alpha-router/alpha-router.ts
ewilz: cool, will probably make a PR adding a few functions to the Fraction lib that would've helped me.

src/routers/alpha-router/alpha-router.ts
NoahZinsmeister: so from my perspective, i think it's ok to update the `optimalRatio` according to the new market price of the target pool after the swap we've just tried. i think the obvious case where this makes sense is if increasing the amount swapped always routed the same or more volume through the target pool. if so, we know the `optimalRatio` is going to be at least what the post-swap price is (we all should be able to agree on this case).

however, even when that's not the case, i still think updating is correct. say we do a single round, the swap routes through our target pool, we update the `optimalRatio`, then do the next round (with a higher swap amount). if this round doesn't route through the same pool, we'll have overshot, but because we still update `exchangeRate`, we still have an accurate picture w/r/t the price impact of that trade size, so our next round (targeting the original `optimalRatio` with the new `exchangeRate`, should compensate for this overshoot, and we should be able to iterate toward the optimal solution.

the one important caveat here is that if a swap _doesn't_ use the target pool, we have to be sure to update the optimalRatio back to what it was before the loop, which afaict the current code doesn't do, it only updates it in the case when the pool is used, but then it's never reset. does that make sense?

src/routers/alpha-router/alpha-router.ts
ewilz: 1) i TOTALLY have this reversed. lol.
2) you're correct, bc I had this reversed, I also made this reverse assumption about where the price is moving.

src/routers/alpha-router/alpha-router.ts
ewilz: thanks, what you mention in your last paragraph is definitely a bug, I will fix that.

willpote: Makes sense to me, though yeah currently the code doesn't reset the optimalRatio

Imo it would be simpler to just reset it on every iteration, but I suppose having the code where we don't reset initially could make it converge quicker in the case where the proposed routes always use the target pool, so I'm ok with it

willpote (approved):

PR #7, opened 2021-09-23
Update eip-1559-gas-price-provider.ts

timnugent:
this.blocksToConsider needs to be an int, not quoted string in the eth_feeHistory RPC call - I get a {"code": -32602, "message": "Invalid method parameter(s)."} without this

Bug fix

32602 - Invalid method parameter(s)

Works as expected

- **Other information**:
JSON_RPC_PROVIDER is Alchemy, Node.js v16.9.1


whenmoon (approved):

willpote:
Thanks. I confirmed this works on Infura too.

PR #32, opened 2021-11-05
return SwapToRatioResponse narrowing types + absoluteValue fix

ewilz:
There's a couple tweaks I'm making to smart-order-router. I will collect them here before doing any verison bumps.

(meant to make this draft)

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -376,6 +376,10 @@ export class AlphaRouter
         inputBalance,
         outputBalance
       );
+      if (amountToSwap.equalTo(0)) {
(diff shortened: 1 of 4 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -652,9 +656,12 @@ export class AlphaRouter implements IRouter<AlphaRouterConfig>, ISwapToRatio<Alp
   }
 
   private absoluteValue(fraction: Fraction): Fraction {
-    if (fraction.lessThan(0)) {
```
ewilz, line 655: this is buggy in fraction lib.

willpote (approved):

PR #41, opened 2021-12-03
Construct swap-and-add parameters for SwapRouter.swapAndAddCallParameters()

ewilz:
Changes to swap_and_add. The Calldata builder takes new parameters now that need to be constructed and passed through.
only the calldata for the swap is passed backl
now calldata for both swap AND add are passed back
- **Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,9 @@
   "scripts": {
     "compile-v3-types": "npx typechain --target ethers-v5 --out-dir src/types/v3 './node_modules/@uniswap/?(v3-core|v3-periphery)/artifacts/contracts/**/*.json'",
     "compile-v2-types": "npx typechain --target ethers-v5 --out-dir src/types/v2 './node_modules/@uniswap/?(v2-core|v2-periphery)/build/*UniswapV2*.json'",
+    "compile-router": "npx typechain --target ethers-v5 --out-dir src/types/other './node_modules/@uniswap/swap-router-contracts/artifacts/contracts/SwapRouter02.sol/SwapRouter02.json'",
     "compile-external-types": "npx typechain --target ethers-v5 --out-dir src/types/other 'src/abis/**/*.json'",
-    "build": "run-p compile-v3-types compile-v2-types compile-external-types && run-p build:*",
+    "build": "run-p compile-v3-types compile-v2-types compile-router compile-external-types && run-p build:*",
     "build:main": "tsc -p tsconfig.json",
     "build:module": "tsc -p tsconfig.module.json",
     "fix": "run-s fix:*",
@@ -43,6 +44,7 @@
     "@types/stats-lite": "^2.2.0",
     "@uniswap/default-token-list": "^2.0.0",
     "@uniswap/router-sdk": "^1.0.1",
+    "@uniswap/swap-router-contracts": "1.0.0",
     "@uniswap/token-lists": "^1.0.0-beta.25",
     "@uniswap/v2-core": "^1.0.1",
     "@uniswap/v2-periphery": "^1.1.0-beta.0",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@types/sinon": "^10.0.2",
     "@types/stats-lite": "^2.2.0",
     "@uniswap/default-token-list": "^2.0.0",
-    "@uniswap/router-sdk": "^1.0.1",
+    "@uniswap/router-sdk": "^1.0.2",
     "@uniswap/swap-router-contracts": "1.0.0",
     "@uniswap/token-lists": "^1.0.0-beta.25",
     "@uniswap/v2-core": "^1.0.1",
--- a/src/providers/swap-router-provider.ts
+++ b/src/providers/swap-router-provider.ts
@@ -4,8 +4,8 @@ import { IMulticallProvider } from './multicall-provider';
 import { ProviderConfig } from './provider';
 
 type ApprovalTypes = {
-	tokenInApproval: number;
-	tokenOutApproval: number;
+	approvalTokenIn: number;
+	approvalTokenOut: number;
 }
 
 /**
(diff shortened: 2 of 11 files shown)


src/providers/swap-router-provider.ts
```diff
@@ -0,0 +1,65 @@
...
+			address: SWAP_ROUTER_ADDRESS,
+			contractInterface: SwapRouter02__factory.createInterface(),
+			functionName: 'getApprovalType',
+			functionParams,
+		})
+
+		if (tx.results[0]?.success && tx.results[1]?.success) {
+			const resultTokenIn = tx.results![0]
+			const resultTokenOut = tx.results![1]
+			// TODO: Dealing with return values.
```
ewilz, line 60: halp plz @willpote .

Not sure how to best deal with `SuccessResult<TReturn>`
Errs from foolin around:
```
src/providers/swap-router-provider.ts:65:76 - error TS7053: Element implicitly has an 'any' type because expression of type '0' can't be used to index type 'Number'.
  Property '0' does not exist on type 'Number'.

65    return { approvalTokenIn: resultTokenIn?.result![0]!, approvalTokenOut: resultTokenOut?.result![0]!}
                                                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

package.json
```diff
@@ -42,7 +43,8 @@
     "@types/sinon": "^10.0.2",
     "@types/stats-lite": "^2.2.0",
     "@uniswap/default-token-list": "^2.0.0",
-    "@uniswap/router-sdk": "^1.0.0-beta.7",
+    "@uniswap/router-sdk": "^1.0.0-beta.8",
```
willpote, line 46: I think we are on 1.0.1 now

src/providers/swap-router-provider.ts
```diff
@@ -0,0 +1,77 @@
+import { Currency, CurrencyAmount } from '@uniswap/sdk-core';
+import { SwapRouter02__factory } from '../types/other';
+import { log } from '../util';
+import { IMulticallProvider } from './multicall-provider';
+
+type ApprovalTypes = {
+  approvalTokenIn: number;
```
willpote, line 7: Lets use the enum from RouterSDK instead of number? https://github.com/Uniswap/router-sdk/blob/main/src/approveAndCall.ts#L7-L13

Hopefully its exported by the package

src/routers/alpha-router/alpha-router.ts
```diff
@@ -518,9 +518,14 @@ export class AlphaRouter
         outputBalance.currency,
         TradeType.EXACT_INPUT,
         swapConfig,
-        { ...DEFAULT_CONFIG, ...routingConfig, protocols: [Protocol.V3] } // TODO: Enable V2 once have trade object across v2/v3.
+        { ...DEFAULT_CONFIG, ...routingConfig, protocols: [Protocol.V3] }, // TODO: Enable V2 once have trade object across v2/v3.
```
willpote, line 521: Would be great to add V2 here at some point. It will break a few of your tests but shouldn't be too bad to fix

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1214,19 +1219,51 @@ export class AlphaRouter
...
-      inputTokenPermit,
-    });
+    let methodParameters: MethodParameters
+    if (!!swapAndAddOptions) {
+      const preLiquidityPosition = swapAndAddOptions.preLiquidityPosition
+
+      const finalBalanceTokenIn = swapAndAddOptions.initialBalanceTokenIn.subtract(trade.inputAmount)
+      const finalBalanceTokenOut = swapAndAddOptions.initialBalanceTokenOut.add(trade.outputAmount)
+      const approvalTypes = await this.swapRouterProvider.getApprovalType(finalBalanceTokenIn, finalBalanceTokenOut)
+      const zeroForOne = finalBalanceTokenIn.currency.wrapped.address < finalBalanceTokenOut.currency.wrapped.address
```
willpote, line 1237: `wrapped.sortsBefore`

src/routers/alpha-router/alpha-router.ts
```diff
@@ -606,7 +610,8 @@ export class AlphaRouter
     quoteCurrency: Currency,
     tradeType: TradeType,
     swapConfig?: SwapConfig,
-    partialRoutingConfig: Partial<AlphaRouterConfig> = {}
+    partialRoutingConfig: Partial<AlphaRouterConfig> = {},
+    swapAndAddParameters?: SwapAndAddParameters,
```
willpote, line 614: Hmm I think it would be cleaner to leave route as is, and make routeToRatio responsible for generating the swapAndAdd calldata. Feels like its a cleaner separation of concerns to have IRouter/route purely responsible for swap calldata, and ISwapToRatio/routeToRatio responsible for swap+liquidity calldata.

src/providers/swap-router-provider.ts
ewilz: making some changes to router-sdk to export it 😬

src/routers/alpha-router/alpha-router.ts
ewilz: woops just seeing this. addressing.

test/unit/routers/alpha-router/alpha-router.test.ts
```diff
@@ -1974,6 +2043,69 @@ describe('alpha router', () => {
...
+        const swap = await alphaRouter.routeToRatio(
+          token0Balance,
+          token1Balance,
+          positionPreLiquidity,
+          SWAP_AND_ADD_CONFIG,
+          SWAP_CONFIG,
+          ROUTING_CONFIG
+        );
+
+        if (swap.status == SwapToRatioStatus.SUCCESS) {
```
willpote, line 2085: Nit: Can we check that methodParameters is defined in the response?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1214,20 +1236,60 @@ export class AlphaRouter
     return trade;
   }
 
-  private buildMethodParameters(
+  private async buildMethodParameters(
```
willpote, line 1239: Can we separate these into 2 methods? Maybe: `buildMethodParametersForSwap` `buildMethodParametersForSwapAndAdd`

willpote (approved):

PR #45, opened 2021-12-21
polygon support checkpoint

snreynolds:


src/util/chains.ts
```diff
@@ -1,3 +1,5 @@
+import { Currency, Ether, NativeCurrency, Token } from '@uniswap/sdk-core';
+import { WRAPPED_NATIVE_CURRENCY } from '..';
```
ewilz, line 2: nit: wonder if it would just make sense to be more explicit with where this dependency is coming from for consistency and readability.

src/providers/token-provider.ts
```diff
@@ -305,6 +305,46 @@ export const UNI_ARBITRUM_RINKEBY = new Token(
...
+
+export const DAI_POLYGON_MUMBAI = new Token(
+  ChainId.POLYGON_MUMBAI,
+  '0x001b3b4d0f3714ca98ba10f6042daebf0b1b7b6f',
+  18,
+  'DAI',
+  'Dai Stablecoin'
+);
+
+export const WRAPPED_NATIVE_CURRENCY: { [chainId: number]: Token } = {
```
ewilz, line 342: cool! I forsee abstracting this away potentially in the future into sdk-core but this looks good to me for now. I wonder if it makes sense to make a tech debt issue in sdk-core? Have you thought about this?

test/unit/routers/alpha-router/functions/best-swap-route.test.ts
```diff
@@ -44,22 +44,22 @@ import {
 const v3Route1 = new V3Route(
   [USDC_DAI_LOW, DAI_USDT_LOW, WETH9_USDT_LOW],
   USDC,
-  WETH9[1]
+  WETH9[1]!
```
ewilz, line 47: why does this need an exclamation mark?? @justindomingue is there a better way to address this typescript sensei?

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -53,11 +54,16 @@ export class V3HeuristicGasModelFactory extends IV3GasModelFactory {
...
     poolProvider: IV3PoolProvider,
     token: Token
+    // this is the quoteToken
   ): Promise<IGasModel<V3RouteWithValidQuote>> {
     // If our quote token is WETH, we don't need to convert our gas use to be in terms
     // of the quote token in order to produce a gas adjusted amount.
     // We do return a gas use in USD however, so we still convert to usd.
-    if (token.equals(WETH9[chainId]!)) {
+
+    // POLYGON TODO
```
justindomingue, line 63:
```suggestion
    // TODO: polygon
```
probably a more standard way of doing it

justindomingue: also can remove?

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -70,25 +76,25 @@ export class V3HeuristicGasModelFactory extends IV3GasModelFactory {
...
         gasCostInUSD: CurrencyAmount;
       } => {
-        const { gasCostInEth, gasUse } = this.estimateGas(
+        const { gasCostNativeCurrency, gasUse } = this.estimateGas(
           routeWithValidQuote,
           gasPriceWei,
           chainId
         );
 
-        const ethToken0 = usdPool.token0.address == WETH9[chainId]!.address;
```
ewilz, line 79: (re: my exclamation mark comment) oh I guess we were using exclamation marks before....? I'm not really a typescript extraordinaire yet, but I guess it must be ok?

src/util/chains.ts
```diff
@@ -97,7 +131,52 @@ export const ID_TO_PROVIDER = (id: ChainId): string => {
...
+  equals(other: Currency): boolean {
+    return other.isNative && other.chainId === this.chainId;
+  }
+
+  get wrapped(): Token {
+    if (!isMatic(this.chainId)) throw new Error('Not matic');
+    return WRAPPED_NATIVE_CURRENCY[this.chainId]!;
+  }
+
+  public constructor(chainId: number) {
```
ewilz, line 159: another readability/consistency nit. I feel like we usually place constructor as the first method?

src/util/chains.ts
```diff
@@ -97,7 +131,52 @@ export const ID_TO_PROVIDER = (id: ChainId): string => {
...
+    if (!isMatic(chainId)) throw new Error('Not matic');
+    super(chainId, 18, 'MATIC', 'Polygon Matic');
+  }
+}
+
+export class ExtendedEther extends Ether {
```
ewilz, line 165: can `ExtendedEther` and `MaticNativeCurrency` maybe be consolidated?? It seems like Ether is almost entirely rewritten anyway. Because then the name `ExtendedEther` is being used elswhere even when it could be MATIC. I'm just confused about these two being separate classes both used for MATIC

cli/commands/quote.ts
```diff
@@ -84,13 +90,14 @@ export class Quote extends BaseCommand {
       tokenOutStr,
     ]);
 
+    // if the tokenIn str is 'ETH' or 'MATIC' or NATIVE_CURRENCY_STRING
     const tokenIn: Currency =
-      tokenInStr == 'ETH'
-        ? Ether.onChain(chainId)
+      tokenInStr in NativeCurrencyName
```
moodysalem, line 95: fyi, the interface just treats MATIC as 'ETH' atm

cli/commands/quote.ts
```diff
@@ -84,13 +90,14 @@ export class Quote extends BaseCommand {
...
+        ? ExtendedEther.onChain(chainId)
         : tokenAccessor.getTokenByAddress(tokenInStr)!;
     const tokenOut: Currency =
-      tokenOutStr == 'ETH'
-        ? Ether.onChain(chainId)
+      tokenOutStr in NativeCurrencyName
+        ? ExtendedEther.onChain(chainId)
```
moodysalem, line 100: i refactored interface to call `nativeOnChain` instead of `ExtendedEther.onChain` because it's confusing that method returns a class that isn't an `ExtendedEther` instance

src/providers/on-chain-gas-price-provider.ts
```diff
@@ -8,6 +8,7 @@ const DEFAULT_EIP_1559_SUPPORTED_CHAINS = [
   ChainId.RINKEBY,
   ChainId.ROPSTEN,
   ChainId.GÖRLI,
+  ChainId.POLYGON_MUMBAI,
```
moodysalem, line 11: til it's not on polygon mainnet yet

src/providers/token-provider.ts
```diff
@@ -305,6 +305,46 @@ export const UNI_ARBITRUM_RINKEBY = new Token(
...
+//polygon tokens
+export const WMATIC_POLYGON = new Token(
+  ChainId.POLYGON,
+  '0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270',
+  18,
+  'WMATIC',
+  'Wrapped MATIC'
+);
+
+export const USDC_POLYGON = new Token(
```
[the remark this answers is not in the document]
moodysalem: token lists could maybe help

src/providers/token-provider.ts
```diff
@@ -305,6 +305,62 @@ export const UNI_ARBITRUM_RINKEBY = new Token(
...
+  'Dai Stablecoin'
+);
+
+export const WRAPPED_NATIVE_CURRENCY: { [chainId: number]: Token } = {
+  ...WETH9,
+  [ChainId.POLYGON]: WMATIC_POLYGON,
+  [ChainId.POLYGON_MUMBAI]: WMATIC_POLYGON_MUMBAI,
+};
+
 export const TOKENS: {
```
moodysalem, line 364: bit of a nondescript variable name.. tokens for what?

src/providers/token-provider.ts
```diff
@@ -305,6 +305,62 @@ export const UNI_ARBITRUM_RINKEBY = new Token(
   'Uni token'
 );
 
+//polygon tokens
+export const WMATIC_POLYGON = new Token(
```
moodysalem, line 309: should WETH be here too?

src/providers/v3/static-subgraph-provider.ts
```diff
@@ -110,6 +115,12 @@ const BASES_TO_CHECK_TRADES_AGAINST: ChainTokenList = {
     USDT_OPTIMISTIC_KOVAN,
     USDC_OPTIMISTIC_KOVAN,
   ],
+  [ChainId.POLYGON]: [USDC_POLYGON, WETH9[ChainId.POLYGON]!, WMATIC_POLYGON],
```
moodysalem, line 118: hmmmmm, non-null assertions are forbidden in the interface, usually code smell, fyi.

but also the WETH on polygon is not a WETH9 contract, it's subtly different, it could cause bugs to bundle them, rather than bundling the wrapped native currency and having weth on polygon just be another token like dai

src/util/chains.ts
```diff
@@ -8,6 +10,8 @@ export enum ChainId {
   OPTIMISTIC_KOVAN = 69,
   ARBITRUM_ONE = 42161,
   ARBITRUM_RINKEBY = 421611,
+  POLYGON = 137,
+  POLYGON_MUMBAI = 80001,
 }
 
 export const ID_TO_CHAIN_ID = (id: number): ChainId => {
```
moodysalem, line 17: this feels like something that should be a typeguard function, switch feels unnecessary
https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/util/chains.ts
```diff
@@ -46,8 +54,30 @@ export enum ChainName {
   OPTIMISTIC_KOVAN = 'optimism-kovan',
```
moodysalem, line 54: these match infura according to comments, but then we have multiple rpc urls anyway, strange

test/unit/routers/alpha-router/alpha-router.test.ts
```diff
@@ -144,7 +144,7 @@ describe('alpha router', () => {
     mockMulticallProvider = sinon.createStubInstance(UniswapMulticallProvider);
 
     mockTokenProvider = sinon.createStubInstance(TokenProvider);
-    const mockTokens = [USDC, DAI, WETH9[1], USDT];
+    const mockTokens = [USDC, DAI, WETH9[1]!, USDT];
```
moodysalem, line 147: why do we have to add non-null assertions

src/providers/token-provider.ts
snreynolds: Yeah, good suggestion. I'll add an issue.

src/util/chains.ts
snreynolds: will leave for now and rework when we move to sdk-core

src/util/chains.ts
snreynolds: its cause we use these in the routing api but yeah, we need to probably just use the project id in SOR

src/util/chains.ts
```diff
@@ -1,3 +1,5 @@
+import { Currency, Ether, NativeCurrency, Token } from '@uniswap/sdk-core';
+import { WRAPPED_NATIVE_CURRENCY } from '../../src/providers/token-provider';
```
ewilz, line 2: you change this back to eliminate the error. not sure why this would be a problem though

`import { WRAPPED_NATIVE_CURRENCY } from '..';`

ewilz: also `../src` wasn't needed in that string..but still doesn't fix it.

cli/commands/quote.ts
snreynolds: ok, this is fine cause its just CLI and the rest should be fine with the interface since you're just calling route

src/util/chains.ts
moodysalem: i think i prefer using separate env variables, so we could use an endpoint provided by a layer 2 if we wanted, just thought it's a bit strange the comment that it matches without the reason

PR #51, opened 2022-01-09
Fix cli order for polygon not formatted properly

yabirgb:
Docs updated

- **Other information**:


willpote (approved):
Thanks!

PR #52, opened 2022-01-10
RouteToRatio edgecase fixes + better logging

ewilz:

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -644,7 +644,7 @@ export class AlphaRouter
       );
 
       if (exchangeRate.equalTo(0)) {
-        log.info('exchangeRate to 0')
+        log.info('exchangeRate to 0');
         return {
           status: SwapToRatioStatus.NO_ROUTE_FOUND,
           error: 'insufficient liquidity to swap to optimal ratio',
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -630,12 +630,26 @@ export class AlphaRouter
...
+          newRatio: newRatio.asFraction.toFixed(18),
+          inputBalanceUpdated: inputBalanceUpdated.asFraction.toFixed(18),
+          outputBalanceUpdated: outputBalanceUpdated.asFraction.toFixed(18),
+          ratioErrorTolerance: swapAndAddConfig.ratioErrorTolerance.toFixed(18),
+          iterationN: n.toString(),
+        },
+        'QuoteToRatio Iteration Parameters'
+      );
+
+      if (exchangeRate.equalTo(0)) {
```
ewilz, line 646: another option other than erroring is to trade what we do have, then add that as liquidity and keep the rest as dust. Alternatively, if the UI gets this error it could suggest trying to swap a smaller amount.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -630,12 +630,26 @@ export class AlphaRouter
...
+        log.info('exchangeRate to 0');
+        return {
+          status: SwapToRatioStatus.NO_ROUTE_FOUND,
```
ewilz, line 649: bc of math, without catching this, we'd return the error on line 562, which is misleading.

willpote (approved):

PR #60, opened 2022-02-01
chore: update providers import from @ethersproject

ianlapham:
- This PR updates the imports for types on `AlphaRouterParams`

- currently, `providers` is imported from ethers

- in this change, we import `providers` from `'@ethersproject/providers'`

- Currenlty in other applications that use the SOR package, dependencies need to be built to create a local SOR. However, some bundles like widgets can not directly import from `ethers`, so updating this dependency would allow for both to work together


ianlapham:
Not needed - use version 5.4.0 of ethersproject/providers to fix

Issue #66, opened 2022-02-17
Unable to validate token SHINJA 


rileyg98:
- **I'm submitting a ...**
  [x] bug report
  [ ] feature request
  [ ] question about the decisions made in the repository
  [ ] question about how to use this project

- **Summary**
Upon trying to calculate a route to swap WETH to Holdenomics or Para Token, we have found that in the routers attempts to find a route, it fails to validate the SHINJA contract during route calculation and fails to return a route. 
- **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)
I've attached a debug trace of the swap with holdenomics - the same occurs on Para.
[debug.txt](https://github.com/Uniswap/smart-order-router/files/8087444/debug.txt)

rileyg98:
I have determined that both tokens have currently got a 0% buy tax, and this may be the common link.

rileyg98:
It appears this is related to #59 - shinja fails to validate, so it considers the token we're swapping for invalid (as opposed to just the shinja pair)

rileyg98:
After further testing it appears if a possible route is not validated successfully, the whole pair is thrown out. Shinja appears to need too much gas in this instance to run a loan or swap. 

After upping the gas per validate to 600k, we successfully get a quote.

willpote:
Thank you for diving into this. I put in a fix here: https://github.com/Uniswap/smart-order-router/commit/c01f48a5b785a9aa4910e37002156f12adaf7e97

And Shinja can now be swapped

PR #67, opened 2022-02-17
Band-aid fix to resolve validator issues causing active error on interface

rileyg98:
Band-aid fix to the gas limit for verification of a pool swap, to allow shinja (a FOT token) to be successfully validated as a potential hop. 

#66 causes the entire pool to be discarded instead of just the pair impacted

Uses 400k gas instead of 300k to validate the swap.

- **Other information**:
This is an emergency fix to resolve the issue being had by multiple tokens. A proper fix to the validation logic will be required.


rileyg98:
It appears 400k fixes one token, but 600k fixes both. 

Do we need a gas limit here, given the router finds the best path anyway? I've upped it to 1 million in my fork.

PR #70, opened 2022-02-22
Arbitrum l1 gas estimates

snreynolds:
Changes
- Updated provider to fetching the on chain gas data from arbitrum and abstracted into a general L2 data fetching provider
- Moved l1 fee such that its only applied to a route once (in best-swap-route)
    - Pass gasModel to the best-swap-route and call `calculateL1GasFees` which now exists on the gas model interface
- Update the quoteGasAdjusted and estimatedGasUsedUSD based on the l1 fee call


src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -429,18 +480,31 @@ export function getBestSwapRouteBy(
...
+  if (tradeType == TradeType.EXACT_INPUT) {
+    const quoteGasAdjustedForL1 = quote.subtract(
+      gasCostsL1.gasCostL1QuoteToken
+    );
+    quoteGasAdjusted = quoteGasAdjustedForL1;
+  } else {
+    const quoteGasAdjustedForL1 = quote.add(gasCostsL1.gasCostL1QuoteToken);
+    quoteGasAdjusted = quoteGasAdjustedForL1;
+  }
+
```
willpote, line 509: So here we are taking the raw quote for each route (not the gas adjusted quote for each route) and adjusting that with the L1 fee.

Don't we need to adjust `quoteGasAdjusted` so that we also include the regular L2 gas costs as well in our total?

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -371,31 +378,58 @@ export function getBestSwapRouteBy(
...
+    // to an arbitrary stable.
     throw new Error(
       `Could not find a USD token for computing gas costs on ${chainId}`
     );
   }
   const usdToken = usdGasTokensByChain[chainId]![0]!;
   const usdTokenDecimals = usdToken.decimals;
 
+  // if on L2, calculate the L1 security fee
+  let gasCostsL1: L2ToL1GasCosts = {
```
willpote, line 410: Nit: gasCostsL1ToL2

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -54,36 +55,112 @@ export class V3HeuristicGasModelFactory extends IV3GasModelFactory {
...
+      let l1Used = BigNumber.from(0);
+      let l1FeeInWei = BigNumber.from(0);
+      if (chainId == ChainId.OPTIMISM) {
+        [l1Used, l1FeeInWei] = this.calculateOptimismToL1SecurityFee(
+          route,
+          swapOptions,
+          l2GasData as OptimismGasData
+        );
+      } else if (chainId == ChainId.ARBITRUM_ONE) {
+        // the fee returned still needs to be multiplied by `gasPriceWei`
```
willpote, line 93: Not the case anymore I think?

willpote (commented):
So far

willpote (commented):
One other thing we need to do is apply the L1 adjustments when we're comparing all the different route combinations here: https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/functions/best-swap-route.ts#L334-L335

right now we just do a sum of gas adjusted

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -334,13 +342,40 @@ export function getBestSwapRouteBy(
...
+              throw new Error("Can't compute L1 gas fees.");
+            } else {
+              const gasCostL1 = await gasModel.calculateL1GasFees!(
+                curRoutesNew as V3RouteWithValidQuote[]
+              );
+              gasCostL1QuoteToken = gasCostL1.gasCostL1QuoteToken;
+            }
+          }
+
+          const tradeType = curRoutesNew[0]?.tradeType;
```
willpote, line 365: Nit: we have this as `routeType` param

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -429,18 +510,30 @@ export function getBestSwapRouteBy(
...
-  );
+  ).add(gasCostL1QuoteToken);
 
-  const quote = sumFn(
+  let quote = sumFn(
     _.map(bestSwap, (routeWithValidQuote) => routeWithValidQuote.quote)
   );
 
+  // Adjust the quoteGasAdjusted for the l1 fee
+  const tradeType = bestSwap[0]?.tradeType;
```
willpote, line 527: Same

willpote (approved):
2 nits, but looks good to go!

Issue #68, opened 2022-02-18
EIP1559GasPriceProvider: doesn't work with hardhat node (mainnet fork)


sblOWPCKCR:
I'm submitting a bug report. 

**Summary**
When trying to run smart-order-router agains local mainnet fork, I get this error:
```
  code: 'SERVER_ERROR',
  body: '{"jsonrpc":"2.0","id":52,"error":{"code":-32602,"message":"Errors encountered in param 0: Invalid value 4 supplied to : QUANTITY"}}',
  error: Error: Errors encountered in param 0: Invalid value 4 supplied to : QUANTITY
      at getResult (/snapshot/liquidator/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:142:21)
      at processJsonFunc (/snapshot/liquidator/node_modules/@ethersproject/web/lib/index.js:344:22)
      at /snapshot/liquidator/node_modules/@ethersproject/web/lib/index.js:276:46
      at step (/snapshot/liquidator/node_modules/@ethersproject/web/lib/index.js:33:23)
      at Object.next (/snapshot/liquidator/node_modules/@ethersproject/web/lib/index.js:14:53)
      at fulfilled (/snapshot/liquidator/node_modules/@ethersproject/web/lib/index.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:96:5) {
    code: -32602,
    data: undefined
  },
  requestBody: '{"method":"eth_feeHistory","params":[4,"latest",[50]],"id":52,"jsonrpc":"2.0"}',
  requestMethod: 'POST',
  url: 'http://127.0.0.1:8550/'
}
```

**Other information**
According to [this spec](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false), 1st argument of `eth_feeHistory` needs to be hex-encoded

Similar issue: https://github.com/ChainSafe/web3.js/issues/4510

We worked around this by forking EIP1559GasPriceProvider and changing this:
```typescript
      const feeHistoryRaw = (await this.provider.send('eth_feeHistory', [
        this.blocksToConsider,
        'latest',
        [this.priorityFeePercentile],
[...]

PR #76, opened 2022-03-17
Resolve `eth_feeHistory` -> `blockCount` argument problem

riordant:
Bug fix

https://github.com/Uniswap/smart-order-router/issues/68

The issue is that the first argument to `eth_feeHistory` (`blockCount`) must be a hex-encoded integer. In the code it is just a `number`. Regardless, it appears to work for most nodes; it however fails in Hardhat (specifically mainnet fork mode). In these cases it will fail with the following error:
`Errors encountered in param 0: Invalid value 4 supplied to : QUANTITY`
in the `getGasPrice()` call.

Works across all nodes due to correctly formatting argument. 

- **Other information**:
-


witoldpilecki88:
Is there a workaround available?

Simply running `npm install --save riordant/smart-order-router` results in the following error at runtime: 

```
auto-router.ts:2:38 - error TS2307: Cannot find module '@uniswap/smart-order-router' or its corresponding type declarations
```

PR #77, opened 2022-03-22
Failing test for AlphaRouter.calculateOptimalRatio()

eliotstock:

Failing test

The ratio returned by AlphaRouter.calculateOptimalRatio() is out by about 2% for this test case.

- **Other information**:


ewilz:
can look into this. I'm betting that the fraction is taking into account Price Impact. If you need to trade through the pool you're providing liquidity to, you will move the price, and hence affect the ratio.

eliotstock:
Thanks for taking a look. I would expect only a tiny price impact for this size of swap in a real pool, but I don't know how the tests handle liquidity in the pool. Isn't the implementation of `calculateOptimalRatio()` completely independent of liquidity in the pool?

PR #90, opened 2022-05-19
Update addresses.ts

a-straus:
This is the old address and might be confusing.  As far as I can see it is not referenced anywhere, and even the swap router provider https://github.com/Uniswap/smart-order-router/blob/9e9987d44348c101b07c976c891d980ba8e89f91/src/providers/swap-router-provider.ts#L12 is using the new address

- const update

- There is none, this address is not being used

- Just an update of a const

- **Other information**:


zhongeric:
Good catch, would you mind removing the local declaration of `SWAP_ROUTER_ADDRESS` in `swap-router-provider.ts` and importing from the updated one in `addresses.ts` as well?

zhongeric:
You will have to merge main back into your branch to pass the new status checks too fyi

snreynolds:
Hi, if you'd like to contribute this change please merge main into this branch so we can approve. Otherwise we will edit this with a separate commit. 

Thanks

willpote:
Fixed in https://github.com/Uniswap/smart-order-router/commit/ef77ef73579fa28a0e9c52d30145951735133911

Thanks though!

Issue #94, opened 2022-05-31
Swapping on route involving non-stablecoin ERC20 exactOut does not result in exact output


zhongeric:
- **I'm submitting a ...**
  [x ] bug report
  [ ] feature request
  [ ] question about the decisions made in the repository
  [ ] question about how to use this project

- **Summary**
routing-api integration tests fail when testing the `eth->erc20 (UNI)` with `exactOut` and through V2 pools. Namely, an exact output is not achieved, rather:
```
      AssertionError: expected '10000.000000000000000171' to equal '10000'
      + expected - actual

      -10000.000000000000000171
      +10000
```

- **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)

This branch on my fork of `routing-api` repo reproduces https://github.com/zhongeric/routing-api/tree/eric/swapped-dust-bug-repro

Additionally, the bug can be seen on my fork of this repo SOR: https://github.com/zhongeric/smart-order-router/tree/eric/refactor-integ-tests

PR #92, opened 2022-05-25
Moving integration tests from routing-api into SOR

zhongeric:
- **What kind of change does this PR introduce?** refactoring

- **What is the current behavior?** integration tests for SOR are done in routing-api

- **What is the new behavior (if this is a feature change)?** integration tests will be internal to SOR, will not need to deploy API to test against it 

- **Other information**:

**Still very much WIP but completed:**
- Integrated hardhat jest plugin to support HH alongside jest
- Added basic integration tests for ERC20->ERC20 and ERC20->ETH (wETH)
- `executeSwaps` works now for erc20-erc20 base test
- Fixed weird v3QuoteProvider unknown error by bumping ethers to 5.6.1
- All integration tests pass & unit tests
- All tests for non mainnet chains pass (except optimisim but those fail on router-api too)

Landing onto eric/refactor-integ-tests

PR #95, opened 2022-06-06
Moving integration tests from routing-api into SOR (#92)

zhongeric:
* refactor test-utils and add integ folder

* Add hardhat^2.6.8

* Add @nomiclabs/hardhat-ethers

* Add ERC20 external types

* trying to figure out hardhat import errors

* remove adjustments to tsconfig.json

* Add mising imports

* finally got integration tests to work with plugin@1.00

* Update plugin to 1.0.1 and fix approve param typing

* Add erc20-eth test, simulated execute swaps still broken

* Clean up code

* remove unused forkAndFund

* Add new jest-environment-hardhat plugin and confirm integ tests pass

* Add erc20-eth large trade and confirm routeString test passes

* Add more checks

* Change slippage to match CLI use

* Update forking blockNumber, fix executeSwaps problem, introduce routing exact out error

* Fix weird v3 quote provider issue by bumping ethers to 5.6.1

* Fix token ordering issue, remove v3poolprovider stuff from route processing

* Pass all tests :)

* Fix out of gas error with hardcoded blockGasLimit in hardhat config

* Finish implementing all of the mainnet tests

* Remove console.logs in quote provider

* Condense last few tests & all tests pass!

* scratch that one test is failing, maybe due to some ERC20 token dust?

* Add comments

* Fix lodash lib error

* remove unused file

* fix local var problem

* Add integ tests for non mainnet chains, all pass except optimism

* Fix native test for non mainnet with nativeOnChain

* Drop scoped jest-env-hardhat for new version 1.1.4

* skip failing test due to known error https://github.com/Uniswap/smart-order-router/issues/94

* remove console.logs and comments

* Remove unused file, clean up comments

* Revisions - generalize helper functions and use them more

* reformat quoteProvider and alpha-router with prettier and address more integ tests revisions

* Fix broken test

Co-authored-by: Eric Zhong <eric.zhong@uniswap.org>

- **Other information**:


test/test-util/schema.ts
```diff
@@ -0,0 +1,80 @@
...
+  amountIn?: string;
+  amountOut?: string;
+};
+
+export type V2Reserve = {
+  token: TokenInRoute;
+  quotient: string;
+};
+
+export type QuoteResponse = {
```
zhongeric, line 49: This type is no longer used (in favor of more concise `

zhongeric (commented):
pushing up some very minor changes now

zhongeric:
Ready for review :)

hardhat.config.js
```diff
@@ -0,0 +1,23 @@
+require('@nomiclabs/hardhat-ethers');
+
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const mainnetFork = {
+  url: `${process.env.JSON_RPC_PROVIDER}`,
+  blockNumber: 14390000,
```
ewilz, line 9: wondering why we have blockNumber in the hardhat config? I'd think we'd always want to run the tests against the latest block

src/providers/eip-1559-gas-price-provider.ts
```diff
@@ -43,7 +43,11 @@ export class EIP1559GasPriceProvider extends IGasPriceProvider {
 
   public async getGasPrice(): Promise<GasPrice> {
     const feeHistoryRaw = (await this.provider.send('eth_feeHistory', [
-      this.blocksToConsider,
+      /**
+       * @note Use BigNumber.from(this.blocksToConsider).toHexString() after hardhat adds support
+       * @see https://github.com/NomicFoundation/hardhat/issues/1585 .___.
+       */
+      `0x${this.blocksToConsider}`,
```
ewilz, line 50: I think this would break if blocksToConsider is >= 10?? (0x10 = 16) ? A little confused as to why we have to do this, but I'm down ?

zhongeric: Ah facepalm yeah that won't work

Going to opt for `BigNumber.from(this.blocksToConsider).toHexString().replace('0x0', '0x')` since it looks like the issue still exists: https://github.com/NomicFoundation/hardhat/issues/1585

hardhat.config.js
zhongeric: In the routing-api we make a call to the deployed api to grab the latest block and subtract like 10 and use that:
https://github.com/Uniswap/routing-api/blob/98439fa9697617da7156cac485ae29747291447f/test/integ/quote.test.ts#L129-L144

I think by default hardhat forks from the latest block which we could use, though aren't there benefits from pinning a block (consistent state, caching, etc.)?

hardhat.config.js
```diff
@@ -0,0 +1,23 @@
...
+};
+
+module.exports = {
+  networks: {
+    hardhat: {
+      blockGasLimit: 150_000_000,
```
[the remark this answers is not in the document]
zhongeric: Hmm yeah this was used for debugging something earlier, I think we can remove.

src/abis/erc20.json
```diff
@@ -0,0 +1,222 @@
+[
```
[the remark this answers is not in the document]
zhongeric: It's used in `test/test-util/getBalanceAndApprove`, and I'm not sure where this is exposed by the sdk

hardhat.config.js
ewilz: IMO I think there's benefits to also not pinning a block, being we're constantly testing against different (and more up-to-date) state. Which battletests the code even more (almost like fuzzing) could eventually catch an edge case. I

Should/are we continuing to do it like the routing-api was?

zhongeric: I agree - and yeah I think the idea is to do things like routing-api. Let's go with forking from the latest block - 10  which is the default config

hardhat.config.js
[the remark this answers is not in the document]
zhongeric: Removed in latest commit

hardhat.config.js
ewilz: I think like consistent state is important for a lot of tests, but these integration tests are the final frontier before prod, so I kind of like that the blockchain state can be unpredictable.

src/abis/erc20.json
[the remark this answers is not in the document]
ewilz: woops, sorry forget I said anything 😶

test/test-util/schema.ts
ewilz: hmm...as a reference to how the `routing-api` works? Unless there's a comment, I'd be afraid someone would come along and see this as dead code and delete. Not fully convinced it should be included.

ewilz:
awesome eric!!! I think maybe you've already discussed this, but getting a git workflow in for the integration tests would be bombb

hardhat.config.js
ewilz: ok realizing we're keeping this in the routing-api as well, but I think it's worth it to test against recent-ish blockchain state.

test/test-util/schema.ts
ewilz: ok I see..maybe you could look into typescript utility types to keep it around? using a partial or maybe omit https://www.typescriptlang.org/docs/handbook/utility-types.html

zhongeric:
thank you! yeah, we're working on adding the github secrets for the ENV variables @ewilz

test/test-util/schema.ts
zhongeric: My initial rationale was to keep it both as reference to the routing api and to show what fields can be extracted from the swap response. I'm leaning towards adding a comment marking it is deprecated

src/providers/token-provider.ts
```diff
@@ -466,3 +473,98 @@ export class TokenProvider implements ITokenProvider {
     };
   }
 }
+
+export const UNI_GORLI = new Token(
```
ewilz, line 477: nit: might make sense to fit this in with the other GÖRLI stuff

src/providers/token-provider.ts
```diff
@@ -154,6 +154,13 @@ export const DAI_GÖRLI = new Token(
   'DAI',
   'Dai Stablecoin'
 );
+export const UNI_GORLI = new Token(
```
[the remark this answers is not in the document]
zhongeric: ah yes ty

ewilz:
Sorry if I'm repeating anything already discussed!: 

I think there's a lot of things we took from the routing-api that now goes into the SOR `src` directory. I think anything we've copied over to src can now be imported to the `routing-api` from `smart-order-router` to avoid duplicate code. Seems like everything to do with supported chains and tokens can come from smart-order-router now?? On that note I think DEFAULT_TOKEN_LIST in the routing-api can be imported just like it is in SOR too.

And more controversially, we have so many complex test helpers duplicated. We could potentially export these under the src/utils to avoid more duplicate code. I could really see this being a pain down the line if things need to change. We can always set this aside for later though. (@willpote @snreynolds  any opinions)

zhongeric:
> I think there's a lot of things we took from the routing-api that now goes into the SOR `src` directory.

I don't think we modified `src/` in SOR significantly besides the `token-provider` stuff required to run some of the integ tests

> Seems like everything to do with supported chains and tokens can come from smart-order-router now??

I think this makes sense!

> we have so many complex test helpers duplicated

I agree, especially helpers like `executeSwap` which are unchanged from routing-api, we should really just be importing them from SOR. However it would be hard to import anything test related from SOR into routing-api imo because routing-api is testing the API level above SOR. Maybe the response validation helpers from SOR can be used in routing-api like `validateSwapRoute()` and `validateExecuteSwap()`

zhongeric:
Ah yes I agree we can import ~all of those from SOR into routing-api. I was mainly talking about the logic in routing integ tests around calling the api (handling errors, building params, etc.)

test/test-util/schema.ts
snreynolds: Could leave a comment. but.. QuoteResponse in the API is not deprecated.. You just mean you're not using it in the tests right?

zhongeric: yea its not deprecated, just not being used. Can mark as unsued

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1133 @@
...
+          expect(swap).toBeDefined();
+          expect(swap).not.toBeNull();
+
+          const { quote, methodParameters } = parseSwap(swap!);
+
+          const { route } = getRouteFromSwap(tradeType, swap!);
+
+          expect(route).not.toBeUndefined;
+
+          const amountInEdgesTotal = _(route)
```
snreynolds, line 541: small, but you could also just calculate the edgeAmounts as you're iterating through the pools in `getRouteFromSwap`

test/test-util/schema.ts
zhongeric: now thinking about it, if we're importing more things from SOR into routing-api, we may likely end up importing this type to use as well.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: true, though in this case we're building up the `amountInEdgesTotal` to check that it's equal to `amountIn`, and I think it would be better to keep the `getRouteFromSwap` as just a helper function that transforms the data

snreynolds:
> > However it would be hard to import anything test related from SOR into routing-api imo because routing-api is testing the API level above SOR.
> 
> I'm referring to things like: `test/test-util/getBalanceAndApprove.ts` and from `test/integ/routers/alpha-router/alpha-router.integration.test.ts`: things like: `checkQuoteToken` `parseSwap` `getRouteFromSwap` `validateSwapRoute`
> 
> I could definitely be missing something, but if they're identical, shouldn't it be not too hard?

Yeah, theres definitely some duplicate code we could move over just into SOR and import into routing-api. Esp all the hardcoded token information. I think we left some code-debt because we were making quick changes in routing-api that we didnt want to have to change in sor, but might be a good task to include in the backlog since there is a lot of repetition.

zhongeric:
Integration tests pass on github actions, takes ~30 minutes. We think it's an issue with jest hanging up after the tests pass. Will touch base with zach about it since he mentioned it being an issue before. For now we only run unit tests in the GH actions.

willpote (changes requested):
Looking good! Only issue is the conversion of SDK types into strings. In the routing-api we are making an HTTP request to an API and checking the response, which is why we are comparing strings.

This package is just a library and returns objects from the Uniswap SDK, so we should be testing those natively without doing conversions to string/float.

src/routers/router.ts
```diff
@@ -182,3 +183,10 @@ export abstract class ISwapToRatio<RoutingConfig, SwapAndAddConfig> {
     routingConfig?: RoutingConfig
   ): Promise<SwapToRatioResponse>;
 }
+
+export class StaticGasPriceProvider implements IGasPriceProvider {
+  constructor(private gasPriceWei: BigNumber) {}
+  async getGasPrice(): Promise<GasPrice> {
+    return { gasPriceWei: this.gasPriceWei };
+  }
+}
```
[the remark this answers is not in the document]
zhongeric: er - yes, we still need this because we have a specific test where we specify the gas price in wei and create the alpha router with that specific gasPriceProvider: https://github.com/Uniswap/smart-order-router/pull/95/files#diff-b1a09e5a7156f19225eba6ca4e2e78d0942580ff1118bab472bf0ea3b1294904R926-R933

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1138 @@
...
+
+      routeResponse.push(curRoute);
+    }
+  }
+
+  return {
+    route: routeResponse,
+    routeString: routeAmountsToString(route),
+  };
+};
```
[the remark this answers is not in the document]
zhongeric: Got it, so we would just be iterating over the returned SOR object to perform the same equality checks?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1138 @@
...
+            '1000000'
+          );
+        });
+
+        /**
+         * Skipped @see https://github.com/Uniswap/smart-order-router/issues/94
+         */
+        xit(`eth -> erc20`, async () => {
+          const tokenIn = Ether.onChain(1) as Currency;
+          const tokenOut = UNI_MAINNET;
```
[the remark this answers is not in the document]
zhongeric: Ah that makes sense. I'll add a comment summarizing the reason and close the respective issue in SOR

test/integ/routers/alpha-router/alpha-router.integration.test.ts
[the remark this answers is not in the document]
willpote: We do, but this check is just seeing that the totals from each split route add up to the total quote. So you would just iterate through the returned `route` array, summing the `quote` for each. Then compare to the overall `quote` value that was returned

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1138 @@
...
+   * @param tradeType
+   * @param targetQuoteDecimalsAmount? - if defined, checks that the quoteDecimals is within the range of this +/- acceptableDifference
+   * @param acceptableDifference? - see above
+   */
+  const validateSwapRoute = async (
+    quoteDecimals: string,
+    quoteGasAdjustedDecimals: string,
+    tradeType: TradeType,
+    targetQuoteDecimalsAmount?: number,
+    acceptableDifference?: number
```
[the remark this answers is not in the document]
zhongeric: Interesting note is that it looks like there isn't support for `.lessThanOrEqualTo` and `.greaterThanOrEqualTo` for operations on CurrencyAmount, so I'm doing the negated version of the other operation `(!lessThan)`

test/test-util/schema.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Mirrored types from router-api for use in tests
+ */
+
+import { MethodParameters } from '@uniswap/v3-sdk';
+
+export type TokenInRoute = {
```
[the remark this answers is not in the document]
zhongeric: Yep, after refactoring none of the types in this file are used anymore. Deleting

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1021 @@
...
+};
+
+export function parseDeadline(deadline: number): number {
+  return Math.floor(Date.now() / 1000) + deadline;
+}
+
+/// @dev had trouble with the return type casting to BigintIsh, wonder if there are overflow issues here
+const expandDecimals = (currency: Currency, amount: number): number => {
+  return amount * 10 ** currency.decimals;
+};
```
zhongeric, line 82: I had issues casting the return of this func to `BigintIsh`, which is what is required by `CurrencyAmount.fromRawAmount(<Currency>, BigintIsh)`. This could overflow if somehow we have a huge number * 10 ** 18, but this is a testing env so I don't think it's too big of an issue

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1021 @@
...
+            .map((routeWithValidQuote) =>
+              tradeType == TradeType.EXACT_INPUT
+                ? BigNumber.from(routeWithValidQuote.quote.quotient.toString())
+                : BigNumber.from(routeWithValidQuote.amount.quotient.toString())
+            )
+            .reduce((cur, total) => total.add(cur), BigNumber.from(0));
+          /**
+           * @dev for exactIn, make sure the sum of the quote to every split = total quote for the route
+           * @dev for exactOut, make sure the sum of the amountIn of every split = total amountIn for the route
+           */
```
zhongeric, line 483: Can remove these comments, was thinking out loud

zhongeric:
Passes all integ and unit tests

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -0,0 +1,1013 @@
...
+
+  /**
+   * Function to validate swapRoute data.
+   * @param quote: CurrencyAmount<Currency>
+   * @param quoteGasAdjusted: CurrencyAmount<Currency>
+   * @param tradeType: TradeType
+   * @param targetQuoteDecimalsAmount?: number - if defined, checks that the quoteDecimals is within the range of this +/- acceptableDifference (non inclusive bounds)
+   * @param acceptableDifference?: number - see above
+   */
+  const validateSwapRoute = async (
```
willpote, line 153: Thanks for adding these functions, makes things a lot cleaner

willpote (approved):
Nit: could you add the new command for running the integ tests to the README? Feel free to make this change and merge it in without another approval from me.

Thanks for doing this!! Going to help us a lot as we develop new features

willpote:
Yeah good call out, I think the best long term solution would be to have a `test-utils` package. 

I was thinking about your idea to export stuff from test, but I think that would mean we'd have to export the entire test folder which would increase package size (which is a no go for widget since they need this package to be pretty small). 

I'd be inclined to merge this for now, and made refactoring to a `test-utils` package a separate task in the backlog. Wdyt?

PR #97, opened 2022-06-13
Support for Celo order routing

Jesse-Sawa:
* adds support for Celo order routing


raghava-pamula:
this change requires changes in routing-api, which contains integration tests I want to run before merging this

raghava-pamula:
We also have to update https://github.com/Uniswap/smart-order-router/blob/main/cli/commands/quote.ts#L93 so that the CLI doesn't break

src/routers/alpha-router/alpha-router.ts
```diff
@@ -407,6 +407,15 @@ export class AlphaRouter
...
         case ChainId.ARBITRUM_ONE:
         case ChainId.ARBITRUM_RINKEBY:
+
+        /*
+          TODO understand the reason behind the "out of gas" error we get if we
+          don't have this here due to low gas limit or low gas price
+          The following gas limits seem to be sufficient for transactions on Celo.
+          Optimism & default limits caused the 'out of gas' error
+        */
+        case ChainId.CELO:
```
raghava-pamula, line 417: Recommend changing multicall chunk for the fallback to 2, as I noticed that `./bin/cli quote --tokenIn 0xF194afDf50B03e69Bd7D057c1Aa9e10c9954E4C9 --tokenOut 0x874069Fa1Eb16D44d622F2e0Ca25eeA172369bC1 --amount 1 --exactIn --protocols v3 --recipient 0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B --minSplits 1 --router alpha --chainId 44787 --debug` doesn't work with these settings due to out of gas

src/providers/multicall-uniswap-provider.ts
```diff
@@ -42,13 +34,10 @@ export class UniswapMulticallProvider extends IMulticallProvider<UniswapMultical
...
     protected provider: BaseProvider,
-    protected gasLimitPerCall = 1_000_000,
-    protected multicallAddressOverride = UNISWAP_MULTICALL_ADDRESS
+    protected gasLimitPerCall = 1_000_000
   ) {
     super();
-    const multicallAddress = multicallAddressOverride
-      ? multicallAddressOverride
-      : contractAddressByChain[this.chainId];
+    const multicallAddress = UNISWAP_MULTICALL_ADDRESSES[this.chainId];
```
willpote, line 40: Why have we removed the functionality for overriding the address?

src/providers/token-provider.ts
```diff
@@ -385,6 +385,122 @@ export const WETH_POLYGON_MUMBAI = new Token(
...
+  6,
+  'Dai on moonbeam bridged using Multichain'
+)
+
+export const WBTC_MOONBEAM = new Token(
+  ChainId.MOONBEAM,
+  '0x922D641a426DcFFaeF11680e5358F34d97d112E1',
+  8,
+  'Wrapped BTC bridged using Multichain'
+)
```
willpote, line 502: We should set the symbol for these tokens

src/routers/alpha-router/alpha-router.ts
```diff
@@ -431,6 +431,39 @@ export class AlphaRouter
             }
           );
           break;
+
+        /*
+          TODO understand the reason behind the "out of gas" error we get if we
+          don't have this here due to low gas limit or low gas price
+          The following gas limits seem to be sufficient for transactions on Celo.
+          Optimism & default limits caused the 'out of gas' error
+        */
```
willpote, line 440: Remove?

src/routers/alpha-router/alpha-router.ts
willpote: Is this comment resolved?

src/routers/alpha-router/gas-models/v3/gas-costs.ts
```diff
@@ -18,6 +18,17 @@ export const BASE_SWAP_COST = (id: ChainId): BigNumber => {
     case ChainId.POLYGON:
     case ChainId.POLYGON_MUMBAI:
       return BigNumber.from(2000);
+
+      //TODO determine if sufficient
```
willpote, line 22: Remove?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -957,31 +972,13 @@ describe('quote for other networks', () => {
...
-            tradeType == TradeType.EXACT_INPUT
-              ? parseAmount('1', tokenIn)
-              : parseAmount('1', tokenOut);
+            chain == ChainId.CELO || chain == ChainId.CELO_ALFAJORES
+              ? tradeType == TradeType.EXACT_INPUT
+                ? parseAmount('10', tokenIn)
+                : parseAmount('10', tokenOut)
+              : tradeType == TradeType.EXACT_INPUT
+                ? parseAmount('100', tokenIn)
+                : parseAmount('100', tokenOut)
```
willpote, line 981: Add a comment as to why we do this

src/util/chains.ts
```diff
@@ -274,6 +345,84 @@ class MaticNativeCurrency extends NativeCurrency {
...
+      return nativeCurrency;
+    }
+    throw new Error(`Does not support this chain ${this.chainId}`);
+  }
+
+  public constructor(chainId: number) {
+    if (!isCelo(chainId)) throw new Error('Not celo');
+    super(chainId, 18, 'CELO', 'Celo');
+  }
+}
```
willpote, line 372: I know we discussed it yesterday but I don't see any logic to handle it. 

By default if the user passes in a NativeCurrency object as token in or token out, when we generate calldata we will add additional logic to wrap/unwrap the native currency. If a user passes in the CeloNativeCurrency object into the SOR, how do we stop router-sdk from adding logic to wrap the token?

willpote (commented):
Can you remove WIP from the PR title if its ready for review?

src/providers/multicall-uniswap-provider.ts
Jesse-Sawa: Since the smart-order-router repository maintains the list of supported chain IDs I don't believe it's necessary to have an override here (it's not used anywhere). 

So I figured removing the unused code would make the constructor a bit cleaner. Let me know what you think! Happy to add it back

Jesse-Sawa: I've also considered the prospect of the next chains integrating Uniswap, contributing to the smart-order-router is a pre-req and they'll have to add their unique [chainId: Multicall address] regardless

src/util/chains.ts
raghava-pamula: Actually I think it is fine as is, it doesn't do any harm and it may be useful to have both NativeCurrency object as well as Wrapped Native Currency.

willpote (approved):

willpote (approved):

PR #101, opened 2022-06-22
Add full SOR support for MixedRoutes

zhongeric:
Main areas of concern:
- New `getMixedRouteCandidatePools()` in `get-candidate-pools.ts`, especially the logic for the heuristic implementation.
- Refactor of `V3QuoteProvider` to `OnChainQuoteProvider`, as well as introducing the generic type used to create instances of the class.
- Logic around what we should do for other networks besides mainnet for mxiedRoutes (we currently only create an instance of `MixedRouteQuoteProvider` for mainnet + eth testnets)

Note: package-lock should not change besides the resolved versions of `ethers` and `ethersproject/providers` being ^5.6.1 and ^5.4.0, respectively. I think current diff is from installing a local tarball of router-sdk, but will fix before merging.


src/routers/alpha-router/alpha-router.ts
```diff
@@ -911,14 +999,25 @@ export class AlphaRouter
...
+    );
+
+    console.log(
+      swapRouteRaw?.routes.map((r) => {
+        return {
+          route: routeToString(r.route),
+          quote: r.quote.toExact(),
+          quoteAdjustedForGas: r.quoteAdjustedForGas.toExact(),
+        };
+      })
```
zhongeric, line 1020: will remove later but very useful for testing without turning on all logs

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1427,22 +1735,69 @@ export class AlphaRouter
...
-    if (hasV3Route && hasV2Route) {
+    if (hasMixedRoute && (hasV3Route || hasV2Route)) {
+      if (hasV3Route && hasV2Route) {
+        metric.putMetric(
+          `MixedAndV3AndV2SplitRoute`,
+          1,
+          MetricLoggerUnit.Count
+        );
+        metric.putMetric(
+          `MixedAndV3AndV2SplitRouteForChain${this.chainId}`,
```
zhongeric, line 1759: not the biggest fan of this naming, but it matches the other ones

src/routers/alpha-router/functions/compute-all-routes.ts
```diff
@@ -104,9 +122,103 @@ export function computeAllRoutes<
...
+        : curPool.token0;
+
+      /// prevent round tripping
+      if (
+        _.flatten(
+          currentRoute.map((pool) => [pool.token0.symbol, pool.token1.symbol])
+        ).includes(currentTokenOut.symbol)
+      ) {
+        continue;
+      }
```
zhongeric, line 190: Should this be allowed for mixedRoutes? It's not reusing the same pool, but the same token between pools.

i.e.

WETH -[v3]-> DAI -[v2] -> USDC -[v3]-> WETH -[v2]-> APE

src/routers/alpha-router/gas-models/mixedRoute/on-chain-heuristic-gas-model.ts
```diff
@@ -0,0 +1,448 @@
...
+    token: Token
+    // this is the quoteToken
+  ): Promise<IGasModel<MixedRouteWithValidQuote>> {
+    const usdPool: Pool = await this.getHighestLiquidityUSDPool(
+      chainId,
+      V3poolProvider
+    );
+
+    /// @dev we might not even need this ... might be only for Op and Arb
+    const calculateL1GasFees = async (): Promise<{
```
zhongeric, line 66: I think we can delete this?

src/routers/alpha-router/gas-models/mixedRoute/on-chain-heuristic-gas-model.ts
```diff
@@ -0,0 +1,448 @@
...
+    );
+    /**
+     * A V3 hop was 1:1 for every pool (i.e. if we went from pool A to pool B it is 1 hop)
+     * Similarly with V2
+     *
+     * Naive approach: every v3 pool is a hop, and counted as 80_000 gas
+     *                 every v2 pool is a hop, and counted as 20_000 gas
+     * Does not matter if v2->v3 or v3->v2
+     * - we will test and see if this hold true
+     */
```
zhongeric, line 295: tldr: I will do some simulations on tenderly(?) to test these assumptions and to build a correct gas model

src/providers/mixed-route-quote-provider.ts
```diff
@@ -0,0 +1,845 @@
+import { BigNumber } from '@ethersproject/bignumber';
```
zhongeric, line 1: This file has little to no changes from `v3QuoteProovider`.

Here's a diff between the two (sdiff -s):
```
eric.zhong@UNISWAP-MAC-020 src % sdiff -s providers/mixed-route-quote-provider.ts providers/v3/quote-provider.ts 
import { encodeMixedRouteToPath } from '@uniswap/router-sdk'; |	import { encodeRouteToPath } from '@uniswap/v3-sdk';
import retry from 'async-retry';			      |	import retry, { Options as RetryOptions } from 'async-retry';
import { MixedRoute } from '../routers/router';		      |	import { V3Route } from '../../routers/router';
import { IQuoterV3__factory } from '../types/other/factories/ |	import { IQuoterV2__factory } from '../../types/v3/factories/
import { ChainId, metric, MetricLoggerUnit, QUOTER_V2_ADDRESS |	import { ChainId, metric, MetricLoggerUnit } from '../../util
import { CurrencyAmount } from '../util/amounts';	      |	import { QUOTER_V2_ADDRESS } from '../../util/addresses';
import { log } from '../util/log';			      |	import { CurrencyAmount } from '../../util/amounts';
import { routeToString } from '../util/routes';		      |	import { log } from '../../util/log';
import { Result } from './multicall-provider';		      |	import { routeToString } from '../../util/routes';
import { UniswapMulticallProvider } from './multicall-uniswap |	import { Result } from '../multicall-provider';
import { ProviderConfig } from './provider';		      |	import { UniswapMulticallProvider } from '../multicall-uniswa
							      >	import { ProviderConfig } from '../provider';
 * @dev import shared types and classes from v3QuoteProvider  |	 * A quote for a swap on V3.
import {						      |	export type V3AmountQuote = {
  BatchParams,						      <
  BlockConflictError,					      <
  BlockNumberConfig,					      <
  FailureOverrides,					      <
  ProviderBlockHeaderError,				      <
  ProviderGasError,					      <
  ProviderTimeoutError,					      <
  QuoteRetryOptions,					      <
  SuccessRateError,					      <
} from './v3/quote-provider';				      <
							      <
/**							      <
 * A on chain quote for a mixed route swap (both V2 and V3 po <
 */							      <
export type MixedRouteAmountQuote = {			      <
							      >	export class BlockConflictError extends Error {
							      >	  public name = 'BlockConflictError';
							      >	}
							      >	export class SuccessRateError extends Error {
							      >	  public name = 'SuccessRateError';
							      >	}
							      >
							      >	export class ProviderBlockHeaderError extends Error {
							      >	  public name = 'ProviderBlockHeaderError';
							      >	}
							      >
							      >	export class ProviderTimeoutError extends Error {
							      >	  public name = 'ProviderTimeoutError';
							      >	}
							      >
							      >	/**
							      >	 * This error typically means that the gas used by the multic
							      >	 * exceeded the total call gas limit set by the node provider
							      >	 *
							      >	 * This can be resolved by modifying BatchParams to request f
							      >	 * quotes per call, or to set a lower gas limit per quote.
							      >	 *
							      >	 * @export
							      >	 * @class ProviderGasError
							      >	 */
							      >	export class ProviderGasError extends Error {
							      >	  public name = 'ProviderGasError';
							      >	}
							      >
							      >	export type QuoteRetryOptions = RetryOptions;
							      >
 * The mixed route and a list of quotes for that route.	      |	 * The V3 route and a list of quotes for that route.
export type MixedRouteWithQuotes = [MixedRoute, MixedRouteAmo |	export type V3RouteWithQuotes = [V3Route, V3AmountQuote[]];
 * Provider for getting mixed route quotes on Uniswap.	      |	 * Provider for getting quotes on Uniswap V3.
 * @interface IMixedRouteQuoteProvider			      |	 * @interface IV3QuoteProvider
export interface IMixedRouteQuoteProvider {		      |	export interface IV3QuoteProvider {
   * For every route, gets an exactIn quote for every amount  |	   * For every route, gets an exactIn quotes on V3 for every 
   * @returns For each route returns a MixedRouteWithQuotes o |	   * @returns For each route returns a V3RouteWithQuotes obje
    routes: MixedRoute[],				      |	    routes: V3Route[],
  ): Promise<{						      |	  ): Promise<{ routesWithQuotes: V3RouteWithQuotes[]; blockNu
    routesWithQuotes: MixedRouteWithQuotes[];		      |
    blockNumber: BigNumber;				      |	  /**
  }>;							      |	   * For every route, gets ane exactOut quote on V3 for every
							      >	   *
							      >	   * @param amountOuts The amounts to get quotes for.
							      >	   * @param routes The routes to get quotes for.
							      >	   * @param [providerConfig] The provider config.
							      >	   * @returns For each route returns a V3RouteWithQuotes obje
							      >	   * @returns The blockNumber used when generating the quotes
							      >	   */
							      >	  getQuotesManyExactOut(
							      >	    amountOuts: CurrencyAmount[],
							      >	    routes: V3Route[],
							      >	    providerConfig?: ProviderConfig
							      >	  ): Promise<{ routesWithQuotes: V3RouteWithQuotes[]; blockNu
							      >	/**
							      >	 * The parameters for the multicalls we make.
							      >	 *
							      >	 * It is important to ensure that (gasLimitPerCall * multical
							      >	 *
							      >	 * V3 quotes can consume a lot of gas (if the swap is so larg
							      >	 * number of ticks), so there is a risk of exceeded gas limit
							      >	 */
							      >	export type BatchParams = {
							      >	  /**
							      >	   * The number of quotes to fetch in each multicall.
							      >	   */
							      >	  multicallChunk: number;
							      >	  /**
							      >	   * The maximum call to consume for each quote in the multic
							      >	   */
							      >	  gasLimitPerCall: number;
							      >	  /**
							      >	   * The minimum success rate for all quotes across all multi
							      >	   * If we set our gasLimitPerCall too low it could result in
							      >	   * quotes failing due to out of gas. This parameters will f
							      >	   * in this case.
							      >	   */
							      >	  quoteMinSuccessRate: number;
							      >	};
							      >
							      >	/**
							      >	 * The fallback values for gasLimit and multicallChunk if any
							      >	 *
							      >	 */
							      >
							      >	export type FailureOverrides = {
							      >	  multicallChunk: number;
							      >	  gasLimitOverride: number;
							      >	};
							      >
							      >	export type BlockHeaderFailureOverridesDisabled = { enabled: 
							      >	export type BlockHeaderFailureOverridesEnabled = {
							      >	  enabled: true;
							      >	  // Offset to apply in the case of a block header failure. e
							      >	  rollbackBlockOffset: number;
							      >	  // Number of batch failures due to block header before tryi
							      >	  attemptsBeforeRollback: number;
							      >	};
							      >	export type BlockHeaderFailureOverrides =
							      >	  | BlockHeaderFailureOverridesDisabled
							      >	  | BlockHeaderFailureOverridesEnabled;
							      >
							      >	/**
							      >	 * Config around what block number to query and how to handle
							      >	 */
							      >	export type BlockNumberConfig = {
							      >	  // Applies an offset to the block number specified when fet
							      >	  // Useful for networks where the latest block may not be av
							      >	  baseBlockOffset: number;
							      >	  // Config for handling header not found errors.
							      >	  rollback: BlockHeaderFailureOverrides;
							      >	};
							      >
 * Computes quotes for mixed routes. For mixed routes, quotes |	 * Computes quotes for V3. For V3, quotes are computed on-cha
 * the 'QuoterV3' smart contract. This is because computing q |	 * the 'QuoterV2' smart contract. This is because computing q
 * exceed this limit. Additionally, each quote on chain can c |	 * exceed this limit. Additionally, each quote on V3 can cons
 * @class MixedRouteQuoteProvider			      |	 * @class V3QuoteProvider
export class MixedRouteQuoteProvider implements IMixedRouteQu |	export class V3QuoteProvider implements IV3QuoteProvider {
   * Creates an instance of MixedRouteQuoteProvider.	      |	   * Creates an instance of V3QuoteProvider.
      : undefined; /// for now				      |	      : QUOTER_V2_ADDRESS;
        `No address for Uniswap QuoterV3 Contract on chain id |	        `No address for Uniswap QuoterV2 Contract on chain id
    if (quoterAddress === QUOTER_V2_ADDRESS)		      <
      throw new Error('QuoterV2 is not supported for mixed pa <
							      <
    routes: MixedRoute[],				      |	    routes: V3Route[],
    routesWithQuotes: MixedRouteWithQuotes[];		      |	    routesWithQuotes: V3RouteWithQuotes[];
							      >	  public async getQuotesManyExactOut(
							      >	    amountOuts: CurrencyAmount[],
							      >	    routes: V3Route[],
							      >	    providerConfig?: ProviderConfig
							      >	  ): Promise<{
							      >	    routesWithQuotes: V3RouteWithQuotes[];
							      >	    blockNumber: BigNumber;
							      >	  }> {
							      >	    return this.getQuotesManyData(
							      >	      amountOuts,
							      >	      routes,
							      >	      'quoteExactOutput',
							      >	      providerConfig
							      >	    );
							      >	  }
							      >
    routes: MixedRoute[],				      |	    routes: V3Route[],
    functionName: 'quoteExactInput',			      |	    functionName: 'quoteExactInput' | 'quoteExactOutput',
    routesWithQuotes: MixedRouteWithQuotes[];		      |	    routesWithQuotes: V3RouteWithQuotes[];
        const encodedRoute = encodeMixedRouteToPath(	      |	        const encodedRoute = encodeRouteToPath(
          false // we don't support exactOut for mixedRoute p |	          functionName == 'quoteExactOutput' // For exactOut 
                    contractInterface: IQuoterV3__factory.cre |	                    contractInterface: IQuoterV2__factory.cre
      .flatMap((routeWithQuotes: MixedRouteWithQuotes) => rou |	      .flatMap((routeWithQuotes: V3RouteWithQuotes) => routeW
    routes: MixedRoute[],				      |	    routes: V3Route[],
  ): MixedRouteWithQuotes[] {				      |	  ): V3RouteWithQuotes[] {
    const routesQuotes: MixedRouteWithQuotes[] = [];	      |	    const routesQuotes: V3RouteWithQuotes[] = [];
      const quotes: MixedRouteAmountQuote[] = _.map(	      |	      const quotes: V3AmountQuote[] = _.map(
        `Failed quotes for mixed routes Part ${idx}/${Math.ce |	        `Failed quotes for V3 routes Part ${idx}/${Math.ceil(
eric.zhong@UNISWAP-MAC-020 src % 
```

src/routers/alpha-router/gas-models/mixedRoute/on-chain-heuristic-gas-model.ts
zhongeric: deleted

src/routers/alpha-router/alpha-router.ts
```diff
@@ -300,6 +314,11 @@ export type AlphaRouterConfig = {
    * This parameter should always be false. It is only included for testing purposes.
    */
   forceCrossProtocol: boolean;
+  /**
+   * Prevent the alpha router from considering mixedRoutes as a valid swap.
+   * Default will be falsy.
+   */
+  disableMixedRoutesConsideration?: boolean;
```
zhongeric, line 321: I was debating adding a `forceMixedRoutes` key here which would force the SOR to return a mixedRoute quote. Would be helpful for testing but not much else. Any preference to including it or not?

src/providers/on-chain-quote-provider.ts
```diff
@@ -270,15 +287,19 @@ export class V3QuoteProvider implements IV3QuoteProvider {
       baseBlockOffset: 0,
       rollback: { enabled: false },
     },
+    protected isMixedRouteQuoteProvider: boolean = false,
```
zhongeric, line 290: Chose to add this extra parameter here to allow us to toggle between which lookup table we use for the Quoter contract addresses, since we can't use the generic type `TRoute` as it's only available at runtime.

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
```diff
@@ -0,0 +1,386 @@
...
+        }
+        acc.push(section);
+        /// we just added a complete v2 section
+        baseGasUse = baseGasUse.add(BASE_SWAP_COST_V2);
+        baseGasUse = baseGasUse.add(
+          /// same behavior in v2 heuristic gas model factory
+          COST_PER_EXTRA_HOP_V2.mul(section.length - 1)
+        );
+      }
+    }
```
zhongeric, line 256: This extra logic for dealing with v2 gas in addition to v3 stuff is the only real difference (besides typing) from the v3 gas model. Wonder if it's worth exploring a refactor.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -459,6 +478,46 @@ export class AlphaRouter
...
+        case ChainId.OPTIMISM:
+        case ChainId.OPTIMISTIC_KOVAN:
+          throw new Error(
+            `Mixed route quote provider not supported for ${chainId}`
+          );
+        case ChainId.ARBITRUM_ONE:
+        case ChainId.ARBITRUM_RINKEBY:
+          throw new Error(
+            `Mixed route quote provider not supported for ${chainId}`
+          );
```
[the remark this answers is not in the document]
zhongeric: And what about for testnets? Or, should we just allow creation for all and it will just naturally not find any routes

src/providers/on-chain-quote-provider.ts
```diff
@@ -319,13 +340,42 @@ export class V3QuoteProvider implements IV3QuoteProvider {
...
+
+    // we only support getting mixedRouteQuotes for chains that support V2 liq
+    if (
+      isMixedRoutes &&
+      !V2_SUPPORTED.some((chainId) => chainId === this.chainId)
+    ) {
+      throw new Error(
+        `Cannot get MixedRoute quotes on ${this.chainId} because it does not support V2 liquidity`
+      );
+    }
```
zhongeric, line 377: May not be necessary since we don't construct an `onChainQuoteProvider<MixedRoute>` for non V2_SUPPORTED chains but extra validation guard in case this class is exported and used as standalone

src/routers/alpha-router/alpha-router.ts
zhongeric: TODO: add logic for passing in Protocol.MIXED to protocols in alphaRouter config forces route through mixedRoutes

zhongeric: just use protocols flag instead

snreynolds (changes requested):

src/providers/on-chain-quote-provider.ts
```diff
@@ -106,49 +114,56 @@ type QuoteBatchPending = {
...
    * @param routes The routes to get quotes for.
    * @param [providerConfig] The provider config.
-   * @returns For each route returns a V3RouteWithQuotes object that contains all the quotes.
+   * @returns For each route returns a RouteWithQuotes object that contains all the quotes.
    * @returns The blockNumber used when generating the quotes.
    */
   getQuotesManyExactIn(
     amountIns: CurrencyAmount[],
-    routes: V3Route[],
+    routes: Route[],
```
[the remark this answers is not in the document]
snreynolds: Ignore if we replace with IRoute

src/routers/alpha-router/alpha-router.ts
snreynolds: one note i think that even if Protocol.MIXED is not specified (as in its just v3 / v2) we should still search for mixed but if mixed is the only flag then just look at mixed

snreynolds (changes requested):

src/providers/on-chain-quote-provider.ts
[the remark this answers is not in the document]
zhongeric: resolved, initially a typo

src/providers/on-chain-quote-provider.ts
```diff
@@ -319,13 +341,39 @@ export class V3QuoteProvider implements IV3QuoteProvider {
...
-    routesWithQuotes: V3RouteWithQuotes[];
+    routesWithQuotes: RouteWithQuotes<TRoute>[];
     blockNumber: BigNumber;
   }> {
+    const isMixedRoutes = routes.every((route) => route instanceof MixedRoute);
+
+    // we cannot have both V3 and MixedRoutes in the same call
+    if (
+      routes.some((route) => route instanceof V3Route) &&
+      routes.some((route) => route instanceof MixedRoute)
```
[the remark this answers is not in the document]
zhongeric: Yep just for external calls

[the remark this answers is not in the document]
zhongeric: Also, we can't use `isMixedRoutes` because we should throw if there are any mixedRoutes/v3routes bundled with others.

src/providers/on-chain-quote-provider.ts
```diff
@@ -319,13 +341,39 @@ export class V3QuoteProvider implements IV3QuoteProvider {
...
+    // we only support getting mixedRouteQuotes for chains that support V2 liq
+    if (
+      isMixedRoutes &&
+      !V2_SUPPORTED.some((chainId) => chainId === this.chainId)
+    ) {
+      throw new Error(
+        `Cannot get MixedRoute quotes on ${this.chainId} because it does not support V2 liquidity`
+      );
+    }
+
```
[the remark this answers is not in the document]
zhongeric: Sg, keeping the `isMixedRoutes` const defined at the parent func level as it is used later on in determining which contract interface to call

src/routers/router.ts
```diff
@@ -21,6 +25,7 @@ import { RouteWithValidQuote } from './alpha-router';
 
 export class V3Route extends V3RouteRaw<Token, Token> {}
 export class V2Route extends V2RouteRaw<Token, Token> {}
+export class MixedRoute extends MixedRouteSDK<Token, Token> {}
```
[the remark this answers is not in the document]
zhongeric: I'm not sure that that would work, as we have some functions in which we construct sdk-routes using the array of pools/pairs, tokenIn, and tokenOut, functionality which is exposed on the SDK route level, not `IRoute`. For example:

```
export function computeAllV3Routes(
  tokenIn: Token,
  tokenOut: Token,
  pools: Pool[],
  maxHops: number
): V3Route[] {
  return computeAllRoutes<Pool, V3Route>(
    tokenIn,
    tokenOut,
    (route: Pool[], tokenIn: Token, tokenOut: Token) => {
      return new V3Route(route, tokenIn, tokenOut);
    },
    pools,
    maxHops
  );
}
```
Where `V3Route` is an extension of the Route in v3-sdk. Whereas the V3 `IRoute` from router-sdk only takes in a precomputed route object.

src/providers/on-chain-quote-provider.ts
```diff
@@ -340,10 +388,13 @@ export class V3QuoteProvider implements IV3QuoteProvider {
 
     const inputs: [string, string][] = _(routes)
       .flatMap((route) => {
-        const encodedRoute = encodeRouteToPath(
-          route,
-          functionName == 'quoteExactOutput' // For exactOut must be true to ensure the routes are reversed.
-        );
+        const encodedRoute =
+          route instanceof V3Route
```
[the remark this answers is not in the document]
zhongeric: See comment below abt using wrapped `IRoute` over sdk routes.

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -941,5 +960,228 @@ export async function getV2CandidatePools({
...
+    .filter((pool) => V2topByTVLPools.map((p) => p.id).includes(pool.id))
+    .sortBy((pool) => -pool.reserveUSD)
+    .value();
+
+  /// we consider all returned V3 pools for this heuristic to "fill in the gaps"
+  const V3sortedPools = _(V3subgraphPools)
+    .sortBy((pool) => -pool.tvlUSD)
+    .value();
+
+  let buildV2Pools: V2SubgraphPool[] = [];
```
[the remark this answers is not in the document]
zhongeric: Yea that thought crossed my mind, however from what I've seen in the results from this heuristic, we add V2 pools to consideration because there is no liquidity on V3 for that token pair much more often than because of them having more USD in the pair. I don't think it's that bad to be a little more loose with selecting V2 pools and tossing them out at the end if they're not favorable since we're not considering too many additional routes rn with `mixedRoutes` being added.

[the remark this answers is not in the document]
zhongeric: I also have no idea how we'd compare liq depth based on price ranges lol, we would likely have to get tick data and then we might as well just calculate the quote off chain

test/test-util/getBalanceAndApprove.ts
```diff
@@ -37,12 +37,6 @@ export const getBalanceAndApprove = async (
     await (
       await aliceTokenIn.approve(approveTarget, constants.MaxUint256)
     ).wait();
-
```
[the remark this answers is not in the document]
zhongeric: Hmm yeah that implementation of approval check is wrong, since for some tokens like USDT, max approval is not uint256 max bc 6 decimals. I wasn't sure if it was the case for other tokens so I just removed it.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -844,6 +922,134 @@ describe('alpha router integration', () => {
...
+        TRIBE_MAINNET
+      );
+      expect(aliceTRIBEBalance).toEqual(parseAmount('10000', TRIBE_MAINNET));
+    });
+
+    describe(`${
+      tradeType === TradeType.EXACT_INPUT ? 'exactInput' : 'exactOutput'
+    } mixedPath routes`, () => {
+      describe('+ simulate swap', () => {
+        it('WISE -> USDC', async () => {
```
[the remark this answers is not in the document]
zhongeric: Those were just for tests I was running locally since I have to create `Token` objects for each token I wanted to test with. I'll remove before merge and clean up the integ tests file

src/routers/alpha-router/alpha-router.ts
zhongeric: Here's what I have implemented:
- if no protocols are specified, we consider V3, V2 like normal, and MIXED only if tradeType is exactInput and we're not disabling it explicitly via config
- else, if we specify MIXED in the protocols array, we will consider them only if the chain supports V2 liquidity and tradeType is correct and config is not set.

Thus, we support the cases where it's `[Protocol.MIXED], [Protocol.V3, Protocol.MIXED]`, etc. However, I'm curious if we want to route mixed if we explicitly do not add V2 into the array.

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
zhongeric: Resolved

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -941,5 +960,232 @@ export async function getV2CandidatePools({
...
+    ...V2candidatePools.selections.topByBaseWithTokenIn,
+    /// tokenOut:
+    ...V2candidatePools.selections.topByTVLUsingTokenOut,
+    ...V2candidatePools.selections.topByBaseWithTokenOut,
+    /// Direct swap:
+    ...V2candidatePools.selections.topByDirectSwapPool,
+  ];
+
+  const V2topByTVLSortedPools = _(V2subgraphPools)
+    .filter((pool) => V2topByTVLPools.map((p) => p.id).includes(pool.id))
```
willpote, line 1029: mapping V2TopByTVLPools is repeated work here. Should precompute it. Also should use a Set<> instead of doing array includes

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -941,5 +960,232 @@ export async function getV2CandidatePools({
...
+    .value();
+
+  /// we consider all returned V3 pools for this heuristic to "fill in the gaps"
+  const V3sortedPools = _(V3subgraphPools)
+    .sortBy((pool) => -pool.tvlUSD)
+    .value();
+
+  /// Finding pools with greater reserveUSD on v2 than tvlUSD on v3, or if there is no v3 liquidity
+  const buildV2Pools: V2SubgraphPool[] = [];
+  V2topByTVLSortedPools.map((V2subgraphPool) => {
```
willpote, line 1040: forEach is more appropriate here than map

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -941,5 +960,232 @@ export async function getV2CandidatePools({
...
+    const V3subgraphPool = V3sortedPools.find(
+      (pool) =>
+        (pool.token0.id == V2subgraphPool.token0.id &&
+          pool.token1.id == V2subgraphPool.token1.id) ||
+        (pool.token0.id == V2subgraphPool.token1.id &&
+          pool.token1.id == V2subgraphPool.token0.id)
+    );
+
+    if (V3subgraphPool) {
+      if (V2subgraphPool.reserveUSD > V3subgraphPool.tvlUSD) {
```
willpote, line 1050: No action here, but just a comment: ideally we would want to compare V2 reserve USD to V3 TVL within +/-% of the current price on V3. Concentrated liquidity makes comparisons like this super weird as all the liquidity could be out of range and thus not being touched during the swap, so TVL can make a V3 pool look better than it really is.

I think this is the best we can do currently and should give good results so no action needed though!

src/routers/alpha-router/gas-models/gas-model.ts
```diff
@@ -78,6 +79,24 @@ export type L1ToL2GasCosts = {
...
 
+export type BuildOnChainGasModelFactoryType = {
+  chainId: ChainId;
+  gasPriceWei: BigNumber;
+  V3poolProvider: IV3PoolProvider;
+  token: Token;
+  l2GasDataProvider?:
+    | IL2GasDataProvider<OptimismGasData>
+    | IL2GasDataProvider<ArbitrumGasData>;
+  V2poolProvider?: IV2PoolProvider;
```
willpote, line 90: Whys this optional? 

Even if its not used in some cases I'd just make it required, and just not use it in the implementation that doesn't need it

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
```diff
@@ -0,0 +1,271 @@
...
+    const nativeV3Pool: Pool | null = await getHighestLiquidityV3NativePool(
+      chainId,
+      token,
+      V3poolProvider
+    );
+
+    let nativeV2Pool: Pair | null;
+    if (V2poolProvider) {
+      /// MixedRoutes
+      nativeV2Pool = await getEthV2Pool(chainId, token, V2poolProvider);
```
willpote, line 119: `getV2NativePool` matches above function better

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
```diff
@@ -0,0 +1,271 @@
...
+          gasCostInToken: CurrencyAmount.fromRawAmount(token, 0),
+          gasCostInUSD: CurrencyAmount.fromRawAmount(usdToken, 0),
+        };
+      }
+
+      /// we will use nativeV2Pool for fallback if nativeV3 does not exist
+      /// can use ! here because we return above if v3Pool and v2Pool are null
+      const nativePool =
+        !nativeV3Pool && nativeV2Pool ? nativeV2Pool : nativeV3Pool!;
+
```
willpote, line 155: The native v3 pool could exist but have no liquidity. Think we should add some sort of heuristic too to check TVL USD is at least some number.. what number I don't know though.. maybe around 500-1000?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -268,6 +274,19 @@ describe('alpha router integration', () => {
...
+        alice
+      );
+    const MixedRouteQuoterV1 = await MixedRouteQuoterV1Factory.deploy(
+      V3_CORE_FACTORY_ADDRESSES[ChainId.MAINNET],
+      V2_FACTORY,
+      WRAPPED_NATIVE_CURRENCY[ChainId.MAINNET].address // TODO: change to be chain dependent
+    );
+
+    const MixedRouteQuoterV1Address = MixedRouteQuoterV1.address;
+
```
willpote, line 289: Why do we deploy instead of using the one you deployed?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -497,6 +543,7 @@ describe('alpha router integration', () => {
         });
 
         it(`eth -> erc20`, async () => {
+          /// Fails for v3 for some reason, ProviderGasError
```
willpote, line 546: hmm?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -844,6 +891,91 @@ describe('alpha router integration', () => {
...
+        [
+          '0xf510dde022a655e7e3189cdf67687e7ffcd80d91', // BOND token whale
+        ]
+      );
+      const aliceBONDBalance = await hardhat.getBalance(
+        alice._address,
+        BOND_MAINNET
+      );
+      expect(aliceBONDBalance).toEqual(parseAmount('10000', BOND_MAINNET));
+    });
```
willpote, line 927: I think we should just add a `forceMixedRoutes` flag for testing instead of this, over time this token combo might start returning v3 only routes or something

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -844,6 +891,91 @@ describe('alpha router integration', () => {
...
+
+    describe(`${
+      tradeType === TradeType.EXACT_INPUT ? 'exactInput' : 'exactOutput'
```
willpote, line 930: We don't support exactOut?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -899,7 +1031,7 @@ describe('quote for other networks', () => {
 
       describe(`${ID_TO_NETWORK_NAME(chain)} ${tradeType} 2xx`, function () {
         // Help with test flakiness by retrying.
-        jest.retryTimes(1);
+        // jest.retryTimes(1);
```
willpote, line 1034: You mean to change this?

src/routers/router.ts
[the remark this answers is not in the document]
zhongeric: Yep, that's about what I had implemented in the PR linked above. I'll req you for review in a bit on that

src/providers/on-chain-quote-provider.ts
```diff
@@ -944,4 +973,34 @@ export class V3QuoteProvider implements IV3QuoteProvider {
...
+  ) {
+    // we cannot have both V3 and MixedRoutes in the same call
+    if (
+      routes.some((route) => route instanceof V3Route) &&
+      routes.some((route) => route instanceof MixedRoute)
+    ) {
+      throw new Error(
+        'Cannot have both V3 and MixedRoutes in the same call to on chain Quoter'
+      );
+    }
```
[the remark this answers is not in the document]
zhongeric: But calls to the existing QuoterV2 will fail if there are mixedRoutes and V3Routes in the same call. You're right though that the new MixedRouteQuoter supports pure v3 routes.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -171,6 +175,10 @@ export type AlphaRouterParams = {
    * The provider for getting V3 quotes.
    */
   v2QuoteProvider?: IV2QuoteProvider;
+  /**
+   * The provider for getting on chain quotes (V3 or MixedRoute)
+   */
+  mixedRouteQuoteProvider?: IOnChainQuoteProvider<MixedRoute>;
```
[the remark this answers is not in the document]
zhongeric: Chatted w/ sara offline abt this, and yeah I agree that we should not have two instances of the `OnChainQuoteProvider` class. I can combine them and push up an implementation soon! However, we thought that enabling getting onChain v2Routes but only for exactInput (because we don't support exactOut at the contract level for mixedRouteQuoter) was a bit weird, because you'd expect to be able to get on chain v2 quotes for exactOut. Wdyt?

src/routers/alpha-router/alpha-router.ts
zhongeric: Gotcha, sounds good - what about the case where no protocols are specified? Should we consider all `v2,v3,mixed`?

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
zhongeric: Also, we chose to make `mixedRouteQuoteProvider` undefined for chains that don't have V2 liquidity via the code here in AutoRouter:
```
        /// @dev We only explicitly support chains with V2 liquidity for mixedRoutes.
        ///      so by default, mixedRouteQuoteProvider is undefined
        case ChainId.RINKEBY:
        case ChainId.ROPSTEN:
        case ChainId.KOVAN:
        case ChainId.GÖRLI:
        case ChainId.MAINNET:
          this.mixedRouteQuoteProvider = new OnChainQuoteProvider(
```
Which I liked for design choice, since it prevented usage of the mixedQuotesProvider for unsupported chains. If we combine the v3 and mixedRoute ones we could throw the error at runtime when quotes are fetched though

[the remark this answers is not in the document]
willpote: > Chatted w/ sara offline abt this, and yeah I agree that we should not have two instances of the `OnChainQuoteProvider` class. I can combine them and push up an implementation soon! However, we thought that enabling getting onChain v2Routes but only for exactInput (because we don't support exactOut at the contract level for mixedRouteQuoter) was a bit weird, because you'd expect to be able to get on chain v2 quotes for exactOut. Wdyt?

Bit odd but I think its ok, just as long as its called out in the docs why this is the case, and you recommend users to use the offchain v2 quote provider instead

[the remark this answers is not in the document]
willpote: > Also, we chose to make `mixedRouteQuoteProvider` undefined for chains that don't have V2 liquidity via the code here in AutoRouter:
> 
> ```
>         /// @dev We only explicitly support chains with V2 liquidity for mixedRoutes.
>         ///      so by default, mixedRouteQuoteProvider is undefined
>         case ChainId.RINKEBY:
>         case ChainId.ROPSTEN:
>         case ChainId.KOVAN:
>         case ChainId.GÖRLI:
>         case ChainId.MAINNET:
>           this.mixedRouteQuoteProvider = new OnChainQuoteProvider(
> ```
> 
> Which I liked for design choice, since it prevented usage of the mixedQuotesProvider for unsupported chains. If we combine the v3 and mixedRoute ones we can throw the `chain not supported` error at runtime when quotes are fetched instead of at construction now

I don't think we need to return an error- we can just not return them the mixed routes even if they explicitly requested it. 

Its similar to V2- if someone specifies V2 as a protocol on a non-V2-supported chain we don't error out we just ignore it and find them the best route in the other protocols they provided. If they specify V2 only on a non-V2 chain, we just return route not found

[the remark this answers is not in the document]
zhongeric: Got it, that pattern makes sense and I think that'll solve some issues I was running into thinking through the implementation

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: Addressed in #158

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: Typo removing in #158

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: Old comment, it doesn't fail anymore - this was before we fixed that block_gasLimit error in hardhatconfig

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: Ah yes we don't need to anymore, removing in #158

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
zhongeric: We don't have access to TVLUSD here, that's only from the subgraph. We can check properties like `Pool.liquidity` or `Pair.getLiquidityValue()`, but I'm hesitant because any number we compare it to is relative depending on the token right?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: ah oops

src/routers/alpha-router/functions/get-candidate-pools.ts
zhongeric: Wdm by precompting it?

zhongeric:
```  
const V2topByTVLSortedPools = _(V2subgraphPools)
    .filter((pool) => V2topByTVLPools.map((p) => p.id).includes(pool.id))
    .sortBy((pool) => -pool.reserveUSD)
    .value();
```
We want all of the `V2SubgraphPools` has type `SubgraphPool[]`, and we want to filter down all of them based on the `PoolId` we selected in building `V2topByTVLPools`. Then, we want to sort them by reserveUSD.

I can change it to this:
```
  const V2topByTVLPools = new Set([
    ...V2candidatePools.selections.topByTVLUsingTokenIn,
    ...V2candidatePools.selections.topByBaseWithTokenIn,
    /// tokenOut:
    ...V2candidatePools.selections.topByTVLUsingTokenOut,
    ...V2candidatePools.selections.topByBaseWithTokenOut,
    /// Direct swap:
    ...V2candidatePools.selections.topByDirectSwapPool,
  ]);

  const V2topByTVLSortedPools = _(V2subgraphPools)
    .filter((pool) =>
      V2topByTVLPools.has({
        id: pool.id,
      })
    )
    .sortBy((pool) => -pool.reserveUSD)
    .value();
```

zhongeric: hmm that actually won't work because Set.has doesn't work like that for objects

zhongeric:
```  
const V2topByTVLPoolIds = new Set(
    [
      ...V2candidatePools.selections.topByTVLUsingTokenIn,
      ...V2candidatePools.selections.topByBaseWithTokenIn,
      /// tokenOut:
      ...V2candidatePools.selections.topByTVLUsingTokenOut,
      ...V2candidatePools.selections.topByBaseWithTokenOut,
      /// Direct swap:
      ...V2candidatePools.selections.topByDirectSwapPool,
    ].map((poolId) => poolId.id)
  );

  const V2topByTVLSortedPools = _(V2subgraphPools)
    .filter((pool) => V2topByTVLPoolIds.has(pool.id))
    .sortBy((pool) => -pool.reserveUSD)
    .value();
```

src/routers/alpha-router/functions/get-candidate-pools.ts
zhongeric: Yeah I used to have it be something like 1.5x more TVL on V2 than V3 but we were missing alot of good pools. Can consider this something to do in the future!

src/routers/alpha-router/gas-models/gas-model.ts
zhongeric: Sounds good

src/providers/on-chain-quote-provider.ts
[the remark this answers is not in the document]
zhongeric: resolved, see changes

src/providers/on-chain-quote-provider.ts
```diff
@@ -944,4 +985,32 @@ export class V3QuoteProvider implements IV3QuoteProvider {
...
+  ) {
+    // we cannot have both V3 and MixedRoutes or V2Routes in the same call since they use different quoter contracts
+    if (
+      !useQuoterV2 &&
+      routes.some((route) => route.protocol === Protocol.V3)
+    ) {
+      throw new Error(
+        'Cannot have mix and match V3 on chain quotes with MixedRoutes and/or V2Routes'
+      );
+    }
```
willpote, line 1008: I still find the comment and logic here super confusing.

If we are using mixedroute quoter, it seems fine to me that some of the routes are V3 only so not sure why we error out.

This would be cleaner to me:

```
if (useQuoterV2 && (routes.some(Protocol.V2) || routes.some(Protocol.MIXED)) {
    error
}
```

src/routers/alpha-router/alpha-router.ts
```diff
@@ -300,6 +309,11 @@ export type AlphaRouterConfig = {
    * This parameter should always be false. It is only included for testing purposes.
    */
   forceCrossProtocol: boolean;
+  /**
+   * Prevent the alpha router from considering mixedRoutes as a valid swap.
```
snreynolds, line 313: Update comment to force the alpha router to route through a mixed route.

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
willpote: Mmm yes thats true.. Ok this is probably fine for now

willpote (approved):

PR #113, opened 2022-07-08
Bump v2 heuristic gas constants

zhongeric:
Testing here vs. tenderly. (green is underestimate compared to simulation, red is over)
Screen Shot 2022-07-08 at 10 18 59 AM

Proposal:
- Increase the V2 base swap fee by 20,000 from 115,000 to 135,000
- Increase the V2 COST_PER_HOP by 30,000 from 20,000 to 50,000

This should bring our heuristic estimates within 5-10% of the actual gas usage (as reported by tenderly) which is much better than currently where we are under by about 30-40%.  However, we'd still be underestimating.

Since it looks like weth-uni is over by 18%, I'm more inclined to keep this update conservative, curious what thoughts are.

### DIFF
--- a/src/routers/alpha-router/gas-models/v2/v2-heuristic-gas-model.ts
+++ b/src/routers/alpha-router/gas-models/v2/v2-heuristic-gas-model.ts
@@ -14,10 +14,10 @@ import {
 } from '../gas-model';
 
 // Constant cost for doing any swap regardless of pools.
-const BASE_SWAP_COST = BigNumber.from(115000);
+const BASE_SWAP_COST = BigNumber.from(135000); // 115000, bumped up by 20_000 @eric 7/8/2022
 
 // Constant per extra hop in the route.
-const COST_PER_EXTRA_HOP = BigNumber.from(20000);
+const COST_PER_EXTRA_HOP = BigNumber.from(50000); // 20000, bumped up by 30_000 @eric 7/8/2022
 
 /**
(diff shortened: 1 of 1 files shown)


snreynolds (approved):

snreynolds:
Would lean on overestimating rather than under.. what does it take to get consistent overestimates?

willpote:
Could you get a sense of how far off the V3 estimates are before merge? Its definitely good to get V2 closer to real world, but we also need to make sure that the V2 and V3 estimates are comparable- i.e. they are approximately the same % off from real world estimates. Otherwise we may send more flow to V3 than we should.

So I guess my question is, are the V3 routes also under by about 30-40% and should we do a similar bump there?

zhongeric:
> So I guess my question is, are the V3 routes also under by about 30-40% and should we do a similar bump there?

I just ran some simulations for various v3 swaps, and currently they are all underestimating around 10%. Notably, swaps involving USDC and/or USDT have a higher gas cost than other tokens (WETH, DAI, etc.).

(green is under estimation, red is over)
Screen Shot 2022-07-08 at 11 34 40 AM

I think it could be fine to leave as is, wdyt?

willpote:
OK cool, looks good to me!

Sara to your point, since WETH-UNI is so far over I'm inclined to leave as is, otherwise we may cause WETH-UNI and other things like that to _wayyy_ over-estimate and not do any splits thus get worse prices

PR #117, opened 2022-07-13
Feat->feat branch add heuristics to mixed routes selection

zhongeric:
/**
     * Apply heuristics to pick the best pools for mixedRoute
     *
     * In order of importance
     * 1. High liq ratio in v2 / v3
     *  - if either tokenIn or tokenOut is substantially more liquid (1.5x?) on V2 than V3 we mark
     * -> done in getCandidatePools
     *
     * 2. Relatively high amountIn / liq in pool
     *
     */


zhongeric (commented):
lgtm

PR #128, opened 2022-07-22
Raghava/tenderly simulation provider

raghava-pamula:
New feature for accurate gas estimates using tenderly simulator

Tenderly provider is given calldata and fallback. It then batch simulates approval+transaction with the calldata and returns the simulation gas_used if request is successful, otherwise logs the incident and returns fallback.
- **Other information**:
Integration test accuracy
[Tenderly overestimates by 11% on average, and up to 24% sometimes](https://docs.google.com/spreadsheets/d/1TyX--YZ9hFHr4YhtmWG-vX_mWJVflW5sh0k3pK3VGL8/edit#gid=1977475015)


src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,102 @@
...
+    } catch(error) {
+      log.info(`Failed to Simulate Via Tenderly!`)
+      if(!fallback) {
+        return new Error('`Failed to Simulate Via Tenderly! No fallback set!`')
+      }
+      log.info(`Defaulting to fallback return value of: ${fallback}s.`)
+      return fallback
+    }
+    log.info({resp:resp.data.simulation_results}, 'Simulated Transaction Via Tenderly')
+    return resp.data.simulation_results[1].transaction.gas_used as number
```
zhongeric, line 100: Can we add some basic typing to the `resp` object so we can check for definedness before evaluating the object? Like, 
```
if(resp.data && resp.data.simulation_results.length > 0) { ... } 
```

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -218,11 +242,22 @@ describe('alpha router integration', () => {
...
+    if(simulate) {
+      simulationGasUsed = await simulateSwap(tokenIn.chainId,tokenIn.wrapped.address,alice._address,methodParameters!)
+      expect(simulationGasUsed instanceof Error).toBe(false)
+    }
+    const { tokenInBefore, tokenInAfter, tokenOutBefore, tokenOutAfter, gasUsed } =
       await executeSwap(methodParameters!, tokenIn, tokenOut!);
+    
+    if(!(simulationGasUsed instanceof Error)) {
+      console.log(`SIMULATED GAS USED: ${simulationGasUsed},\nACTUAL GAS USED: ${gasUsed}`)
+      expect(Math.abs(gasUsed-simulationGasUsed as number)<0.01*gasUsed).toBe(true);
```
zhongeric, line 259: I'd add a comment describing the condition you're testing for here. Like `Ensure that simulated gas is within 1% of actual gas`

Also are you using prettier locally?

zhongeric (approved):

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -103,6 +104,26 @@ describe('alpha router integration', () => {
...
+    fromAddress: string,
+    methodParameters: MethodParameters,
+  ): Promise<number|Error> => {
+    const tp = new TenderlyProvider(
+      process.env.TENDERLY_BASE_URL!,
+      process.env.TENDERLY_USER!,
+      process.env.TENDERLY_PROJECT!,
+      process.env.TENDERLY_ACCESS_KEY!
+    )
+    return await tp.simulateTransaction(
```
zhongeric, line 119: nit: can probably remove this `await` here

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -360,15 +395,17 @@ describe('alpha router integration', () => {
           const { quote, quoteGasAdjusted, methodParameters } = swap!;
 
           await validateSwapRoute(quote, quoteGasAdjusted, tradeType, 100, 10);
-
+          
+          console.log(`TOKEN IN: ${tokenIn.name}\nTOKEN OUT:${tokenOut.name}`)
```
zhongeric, line 399: remove console.logs before merge

test/integ/routers/alpha-router/alpha-router.integration.test.ts
raghava-pamula: done
and yeah I ran `npx prettier --write src/*`

src/providers/tenderly-simulation-provider.ts
raghava-pamula: done

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+    tokenInAddress: string,
+    fromAddress: string,
+    fallback?: number
+  ) => Promise<number|Error>
+}
+export class TenderlyProvider implements ISimulator {
+  TENDERLY_BASE_URL: string
+  TENDERLY_USER: string
+  TENDERLY_PROJECT: string
+  TENDERLY_ACCESS_KEY: string
```
willpote, line 34: Variables within a class should me camelCase. Also set private visibility (default in ts is public)

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+  constructor(TENDERLY_BASE_URL: string, TENDERLY_USER: string, TENDERLY_PROJECT: string, TENDERLY_ACCESS_KEY: string) {
+    this.TENDERLY_BASE_URL = TENDERLY_BASE_URL
+    this.TENDERLY_USER = TENDERLY_USER
+    this.TENDERLY_PROJECT = TENDERLY_PROJECT
+    this.TENDERLY_ACCESS_KEY = TENDERLY_ACCESS_KEY
+  }
+
+  public async simulateTransaction(
+    chainId: number,
+    hexData: string,
```
willpote, line 44: calldata

src/util/callData.ts
```diff
@@ -0,0 +1,6 @@
+//V3 Swap Router Contract
```
willpote, line 1: Space after //

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+    tokenInAddress: string,
```
willpote, line 45: We should take the Token object

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+    fromAddress: string,
+    fallback?: number,
+    stateOverrides?: Record<string, unknown>
+  ): Promise<number|Error> {
```
willpote, line 49: This function should take a SwapRoute object, and return a new SwapRoute object (with all the gas values updated)

It also shouldn't return an Error. On error it should throw.

willpote (changes requested):
This isn't what I meant with my comment [here](https://github.com/Uniswap/routing-api/pull/62#pullrequestreview-1041989826). As implemented we aren't replacing the gas estimates that we return.

The code should be part of the main route() flow. not just have a separate provider like this. 

So when you construct `AlphaRouter` you should pass in a Simulator (it would be optional), 
and if it is provided there is some logic at the end of route that will run the simulator, and overwrite the gas estimates.

So rough list of changes needed:

1/ Update AlphaRouter constructor to take optional ISimulate. Defaults to null
2/ Update SwapOptions to take `from` as am optional parameter
3/ Add `if(this.simulator && swapOptions.from) { ... }` logic to the end of `route()` to run the simulation after the best route is found, and update the SwapRouter object that is returned.

willpote:
Also what about the eth_estimateGas logic that we discussed on Slack?

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+      from: fromAddress,
+      gasPrice: "0",
+      gas: 30000000,
+      type: 1,
+      estimate_gas: true,
+      stateOverrides: stateOverrides
+    }
+
+    const body = {"simulations": [approve, swap]}
+    body
```
willpote, line 84: ?

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+    const opts = {
+      headers: {
+        'X-Access-Key': this.TENDERLY_ACCESS_KEY,
+      },
+    }
+    const url = TENDERLY_BATCH_SIMULATE_API(this.TENDERLY_BASE_URL, this.TENDERLY_USER, this.TENDERLY_PROJECT)
+    const resp=await axios.post(url, body, opts)
```
willpote, line 91: Spacing

raghava-pamula:
> Also what about the eth_estimateGas logic that we discussed on Slack?

The tenderly batch api doesn't support this yet (they're working on it this week). But I think based on the accuracy here: [Tenderly overestimates by 11% on average, and up to 24% sometimes](https://docs.google.com/spreadsheets/d/1TyX--YZ9hFHr4YhtmWG-vX_mWJVflW5sh0k3pK3VGL8/edit#gid=1977475015) , eth_estimateGas isn't needed. Let me know what you think

willpote:
We should also add a flag to the response of route() that indicates if the simulation was run successfully or not. Then the routing API can return that too, and clients will know whether they can trust the gas estimate or not

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,125 @@
...
+      network_id: tokenIn.chainId,
+      input: calldata,
+      to: V3_ROUTER2_ADDRESS,
+      value: "0",
+      from: fromAddress,
+      gasPrice: "0",
+      gas: 30000000,
+      type: 1,
+      estimate_gas: true,
+    }
```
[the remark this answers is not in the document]
willpote: I think the accounts we are using for `fromAddress` in the integ tests may have so much WETH that we don't use the 'value' passed in (we just use their WETH balance). If you write a test where fromAddress has 0 WETH then I expect the simulation will fail for ETH in swaps

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,125 @@
...
+      log.info({approve:resp.data.simulation_results[0],swap:resp.data.simulation_results[1]}, 'Simulated Transaction Via Tenderly')
+
+      // Parse the gas used in the simulation response object, and then pad it so that we overestimate.
+      const gasUsed = resp.data.simulation_results[1].transaction.gas_used * ESTIMATE_MULTIPLIER
+      const l1FeeInWei = BigNumber.from((gasUsed).toFixed())
+      const { gasCostL1QuoteToken, gasCostL1USD } = await getGasCostsInUSDandQuote(tokenIn, l1FeeInWei, poolProvider)
+      // gasUsedL1 is the gas units used calculated from the bytes of the calldata
+      // gasCostL1USD and gasCostL1QuoteToken is the cost of gas in each of those tokens
+      route = {...route, estimatedGasUsed:l1FeeInWei, estimatedGasUsedUSD: gasCostL1USD, estimatedGasUsedQuoteToken: gasCostL1QuoteToken}
+      return route
```
willpote, line 116: What about quoteGasAdjusted

Also this doesn't seem to support exactIn/exactOut correctly

- for exactIn the quote token is tokenOut (i.e. user specifies an amount of token in, auto routers returns their quote for receiving token out)
- for exactOut the quote token is tokenIn (i.e. user specifies an amount of token out, auto router returns their quote for receiving token in)

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,125 @@
...
+ * @export
+ * @interface ISimulator
+ */
+export interface ISimulator {
+  /**
+   * Returns the gas fee that was paid to land the transaction in the simulation.
+   * @returns number or Error
+   */
+  simulateTransaction: (
+    tokenIn: Token,
```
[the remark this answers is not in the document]
raghava-pamula: It is a bit complex to parse tokenIn from swaproute, prefer to keep this as is

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,125 @@
...
+    fromAddress: string,
+    route: SwapRoute,
+    poolProvider: IV3PoolProvider
```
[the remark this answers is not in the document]
raghava-pamula: I can do this, but then there will be two separate instance of v3poolprovider: one belonging to simulator and one belonging to alpha-router. Currently, only one is needed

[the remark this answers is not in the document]
raghava-pamula: unless we pass in the v3poolprovider from simulator to alpha router on construction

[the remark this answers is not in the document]
zhongeric: It's a common pattern in alpha-router that we have an instance of the poolProviders (v3 + v2) on the `AlphaRouter` class and we pass them into the constructors of other classes. See the constructor of `V3RouteWithValidQuote` for how we pass in the poolProvider there.

src/abis/v3SwapRouter.json
```diff
@@ -0,0 +1 @@
...int256","name":"amountMinimum","type":"uint256"},{"internalType":"address","name":"recipient","type":"address"}],"name":"unwrapWETH9","outputs":[],"stateMutability":"payable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amountMinimum","type":"uint256"}],"name":"unwrapWETH9","outputs":[],"stateMutability":"payable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amountMinimum","type":"uint256"},{"internalType":"address","name":"recipient","type":"address"},{"internalType":"uint256","name":"feeBips","type":"uint256"},{"internalType":"address","name":"feeRecipient","type":"address"}],"name":"unwrapWETH9WithFee","outputs":[],"stateMutability":"payable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amountMinimum","type":"uint256"},{"internalType":"uint256","name":"feeBips","type":"uint256"},{"internalType":"address","name":"feeRecipient","type":"address"}],"name":"unwrapWETH9WithFee","outputs":[],"stateMutability":"payable","type":"function"},{"inputs":[{"internalType":"uint256","name":"value","type":"uint256"}],"name":"wrapETH","outputs":[],"stateMutability":"payable","type":"function"},{"stateMutability":"payable","type":"receive"}]
```
zhongeric, line 1: Why do you need to import this ABI? I thought we only use the SwapRouter02

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,259 @@
...
+      from: fromAddress,
+      gasPrice: "0",
+      gas: 30000000,
+    }
+
+    const swap = {
+      network_id: tokenIn.chainId,
+      input: calldata,
+      to: V3_ROUTER2_ADDRESS,
+      value: BigNumber.from(route.methodParameters.value).toString(),
```
willpote, line 225: Pretty sure simulation will fail if the fromAddress does not have this much native eth? 

I think the integ test accounts have huge balances of ETH so the tests are probably passing incorrectly?

willpote: We should only set this if the original token in specified (i.e. the Currency object passed into the top level route() call) is native

willpote (commented):
Getting v close. One thing is youll need to merge main in, Erics made a bunch of change so you may have some merge conflicts to work through.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -859,220 +1389,7 @@ describe('alpha router integration', () => {
       });
     });
   }
-
-  describe('Mixed routes', () => {
```
zhongeric, line 863: Why are these tests deleted? These are in main and should be merged into your feature

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,324 @@
...
+      try {
+        const swapRouteWithGasEstimate = await this.ethEstimateGas(
+          fromAddress,
+          swapRoute,
+          l2GasData
+        );
+        return swapRouteWithGasEstimate;
+      } catch (err) {
+        log.info({ err: err }, 'Error calling eth estimate gas!');
+        return { ...swapRoute, simulationError: false };
```
willpote, line 172: Should be `true`. The simulator class failed to produce an accurate simulated gas estimate and didn't update the swapRoute object, so the flag needs to be set to true

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,324 @@
...
+    l2GasData?: ArbitrumGasData | OptimismGasData
+  ): Promise<SwapRoute> {
+    const currencyIn = swapRoute.trade.inputAmount.currency;
+    const tokenIn = currencyIn.wrapped;
+    const chainId = tokenIn.chainId;
+    if ([ChainId.CELO, ChainId.CELO_ALFAJORES].includes(chainId)) {
+      const msg = 'Celo not supported by Tenderly!';
+      log.info(msg);
+      throw new Error(msg);
+    }
```
willpote, line 225: Commented this in the last review. I think we should return simulationFailed = true for this

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,324 @@
...
+      !resp.simulation_results[1].transaction ||
+      resp.simulation_results[1].transaction.error_message
+    ) {
+      const msg = `Failed to Simulate Via Tenderly!: ${resp.simulation_results[1].transaction.error_message}`;
+      log.info(
+        { err: resp.simulation_results[1].transaction.error_message },
+        msg
+      );
+      throw new Error(msg);
+    }
```
willpote, line 288: Also commented this in the last review. Should definitely return simulationFailed = true for this one.

src/providers/tenderly-simulation-provider.ts
```diff
@@ -0,0 +1,324 @@
...
+ * Provider for dry running transactions.
+ *
+ * @export
+ * @interface ISimulator
+ */
+export interface ISimulator {
+  /**
+   * Returns a new SwapRoute with updated gas estimates
+   * @returns SwapRoute
+   */
```
willpote, line 54: You should mention that all classes that implement must set the simulationFailed field to true if the simulation fails.

It should part of the contract of the interface

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -228,14 +241,25 @@ describe('alpha router integration', () => {
...
+        if(tokenIn.equals(nativeOnChain(1))) {
+          console.log(tokenIn, tokenInBefore.toFixed(3), tokenInAfter.toFixed(3), checkTokenInAmount, estimatedGasUsed)
+          console.log(tokenInBefore
+            .subtract(tokenInAfter)
+            .equalTo(
+              CurrencyAmount.fromRawAmount(
+                tokenIn,
+                expandDecimals(tokenIn, checkTokenInAmount)
+              )
+            ))
```
willpote, line 261: Remove

willpote (approved):

PR #154, opened 2022-08-04
Move hardhat and hardhat-ethers to dev dependencies

zhongeric:
From what I can tell, `hardhat` and `@nomiclabs/hardhat-ethers` are two packages that I added when adding integ tests to SOR that should be in devDependencies. Not sure if this affects the .lock file change for interface but I can test locally.


package.json
```diff
@@ -31,7 +31,6 @@
     "node": ">=10"
   },
   "dependencies": {
-    "@nomiclabs/hardhat-ethers": "^2.0.6",
     "@types/async-retry": "^1.4.2",
```
zzmp, line 34: The entire `@types` scope should probably also be moved to devDependencies - they are only used for compilation.

package.json
```diff
@@ -53,7 +52,6 @@
     "ethers": "^5.6.1",
```
zzmp, line 52: Is bunyan-debug-stream development-only?

zzmp (approved):

zzmp:
This will likely affect the lockfile for interface. If you'd like, I can own updating interface once this is merged and released under a new patch version.

willpote (approved):
Agree with Zacks comments but approved

package.json
zhongeric: Yeah it's just. for logging

zhongeric:
> This will likely affect the lockfile for interface. If you'd like, I can own updating interface once this is merged and released under a new patch version.

That would be great as I might not be here anymore after it's merged haha

zhongeric:
Moved all @types into devDep, had to bump @types/async-retry to 1.4.4 (what it was resolving to anyways) since 1.4.2 was throwing errors when building.

zzmp:
> > This will likely affect the lockfile for interface. If you'd like, I can own updating interface once this is merged and released under a new patch version.
> 
> That would be great as I might not be here anymore after it's merged haha

This is actually taken care of in https://github.com/Uniswap/interface/pull/4326.

PR #155, opened 2022-08-04
feat -> feat Eric/refactor gas factories

zhongeric:
…extend v3Factory class

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -911,7 +911,7 @@ export class AlphaRouter
       chainId: this.chainId,
       gasPriceWei,
       V3poolProvider: this.v3PoolProvider,
-      inTermsOfToken: quoteToken,
+      token: quoteToken,
       l2GasDataProvider: this.l2GasDataProvider,
     });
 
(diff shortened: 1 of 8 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -921,7 +920,6 @@ export class AlphaRouter
         this.chainId,
         gasPriceWei,
         this.v3PoolProvider,
-        this.v2PoolProvider,
```
snreynolds, line 924: Why don't we need this in alpha-router?

snreynolds: resolved

snreynolds (approved):

snreynolds (changes requested):
There is an edgecase where I think we actually do need to pass in a V2PoolProvider. If the quote token in the route is not a native token to the chain, and has better liquidity with a usd pool on v2, that might give us a more accurate price for the token in usd. Or say maybe the quote token doesnt even exist on v3 (only on v2) then we need to get the pair from v2 instead of v3.

snreynolds:
> Decided to use v2 quoteToken/native pool as a fallback if the v3 pool does not exist. Additionally, we could consider their respective TVLs / prices, but I feel like it's not necessary for now since this is just for gas estimations.

I agree. I think the fallback is good

PR #156, opened 2022-08-08
feat -> feat Add protocol flag to SOR version of SDK routes

zhongeric:
Few notes:
- Sara initially suggested that we use the wrapped `IRoute` instead of the SDK routes in SOR to get access to the `protocol` key and the abstracted `pools` and `path` keys.
- I tried that out, importing the `IRoute` classes and wrapping it around the new SDK routes that we create, but it added a lot of complexity to the code that I didn't really think was worth it. Namely, in computing routes and the tests.
- So since the initial goal was to remove the `instanceOf` checks against `V3Route` and `MixedRoute`, I added a protocol key to each class in SOR that we can use to differentiate with.

However I think there's merit in porting over the SOR to use the `IRoutes` entirely from router-sdk, but it's a change that I think can happen after launch


zhongeric:
Need this merged for basePR, merging for now

PR #158, opened 2022-08-09
feat->feat Refactor on chain quote provider

zhongeric:
Feature branch for implementing:
- Using one instance of `OnChainQuoteProvider` for V3, MixedRoute, and V2 quotes.
- Removing the `disableMixedRoutesConsideration` flag and adjusting the logic for protocol queries
- Adding `forceMixedRoutes`
- Adding various tests for `getMixedRouteQuotes()` in alpha-router

TODO:
- need to test that we can call the onChainQuoteProvider and get on chain quotes for V2 exactIn routes

### DIFF
--- a/src/providers/on-chain-quote-provider.ts
+++ b/src/providers/on-chain-quote-provider.ts
@@ -356,7 +356,7 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
       routes.some((route) => route instanceof MixedRoute) ||
       routes.some((route) => route instanceof V2Route);
 
-    /// Validate that there are no incorrect routes / funcitonName combinations
+    /// Validate that there are no incorrect routes / function combinations
     this.validateRoutes(routes, functionName, usesMixedRouteQuoter);
 
     let multicallChunk = this.batchParams.multicallChunk;
@@ -988,7 +988,7 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
   }
 
   /**
-   * Throw an error for incorrect routes / funcitonName combinations
+   * Throw an error for incorrect routes / function combinations
    * @param routes Any combination of V3, V2, and Mixed routes.
    * @param functionName
    * @param usesMixedRouteQuoter Boolean indicating whether the mixedRouteQuoter needs to be used (for pure V2 and Mixed routes)
--- a/src/providers/on-chain-quote-provider.ts
+++ b/src/providers/on-chain-quote-provider.ts
@@ -1,6 +1,10 @@
 import { BigNumber } from '@ethersproject/bignumber';
 import { BaseProvider } from '@ethersproject/providers';
-import { encodeMixedRouteToPath, Protocol } from '@uniswap/router-sdk';
+import {
+  encodeMixedRouteToPath,
+  MixedRouteSDK,
+  Protocol,
+} from '@uniswap/router-sdk';
 import { encodeRouteToPath } from '@uniswap/v3-sdk';
 import retry, { Options as RetryOptions } from 'async-retry';
 import _ from 'lodash';
@@ -9,7 +13,7 @@ import stats from 'stats-lite';
 import { MixedRoute, V2Route, V3Route } from '../routers/router';
 import { IMixedRouteQuoterV1__factory } from '../types/other/factories/IMixedRouteQuoterV1__factory';
 import { IQuoterV2__factory } from '../types/v3/factories/IQuoterV2__factory';
-import { ChainId, metric, MetricLoggerUnit, V2_SUPPORTED } from '../util';
+import { ChainId, metric, MetricLoggerUnit } from '../util';
 import {
   MIXED_ROUTE_QUOTER_V1_ADDRESSES,
   QUOTER_V2_ADDRESSES,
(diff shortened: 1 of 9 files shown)


src/providers/on-chain-quote-provider.ts
```diff
@@ -282,30 +288,32 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
...
     }
-
-    this.quoterAddress = quoterAddress;
+    return quoterAddress;
   }
 
-  public async getQuotesManyExactIn(
+  public async getQuotesManyExactIn<
+    TRoute extends V3Route | V2Route | MixedRoute
+  >(
```
zhongeric, line 311: I like using the generic type here because it lets us type the output, which helps a lot as we call this function to generate the corresponding `RouteWithValidQuote` classes which are distinct depending on the route type.

src/providers/on-chain-quote-provider.ts
```diff
@@ -332,20 +341,23 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
...
-    const isMixedRoutes = routes.every(
-      (route) => route.protocol === Protocol.MIXED
-    );
+    const usesMixedRouteQuoter =
+      routes.some((route) => route instanceof MixedRoute) ||
+      routes.some((route) => route instanceof V2Route);
 
-    this.validateRoutes(routes, functionName, isMixedRoutes);
+    /// Validate that there are no incorrect routes / funcitonName combinations
+    this.validateRoutes(routes, functionName, usesMixedRouteQuoter);
```
zhongeric, line 360: While we will never make a call to `getQuotesManyExactIn()` with a mixture of V3Route | V2Route | MixedRoute through AlphaRouter (since we only use routes created by functions like `computeAllV3Routes`, etc., since this class can be exported we should prevent errors that can arise from incorrect usages.

src/providers/on-chain-quote-provider.ts
```diff
@@ -367,7 +379,11 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
                 route,
                 functionName == 'quoteExactOutput' // For exactOut must be true to ensure the routes are reversed.
               )
-            : encodeMixedRouteToPath(route);
+            : encodeMixedRouteToPath(
+                route instanceof V2Route
+                  ? new MixedRouteSDK(route.pairs, route.input, route.output)
+                  : route
```
zhongeric, line 385: Type wrangling that we have to do in order to support getting V2Routes on chain.

I'm not the biggest fan of this architecture...

src/providers/on-chain-quote-provider.ts
```diff
@@ -971,32 +987,30 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
     }
   }
 
+  /**
+   * Throw an error for incorrect routes / funcitonName combinations
```
willpote, line 991: function

src/providers/on-chain-quote-provider.ts
```diff
@@ -332,20 +341,23 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
...
   }> {
-    const isMixedRoutes = routes.every(
-      (route) => route.protocol === Protocol.MIXED
-    );
+    const usesMixedRouteQuoter =
+      routes.some((route) => route instanceof MixedRoute) ||
+      routes.some((route) => route instanceof V2Route);
 
-    this.validateRoutes(routes, functionName, isMixedRoutes);
+    /// Validate that there are no incorrect routes / funcitonName combinations
```
willpote, line 359: function

src/providers/on-chain-quote-provider.ts
```diff
@@ -971,32 +987,30 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
...
     if (
       routes.some((route) => route instanceof V3Route) &&
-      routes.some((route) => route instanceof MixedRoute)
+      usesMixedRouteQuoter
     ) {
       throw new Error(
-        'Cannot have both V3 and MixedRoutes in the same call to on chain Quoter'
+        'Cannot have mix and match V3 on chain quotes with MixedRoutes and/or V2Routes'
       );
     }
```
willpote, line 1009: Not sure I follow this. In my mind there are just 2 conditions:

1/ 'exactOut' must go on original quoter (which is covered below) 
2/ mixedRoutes must go on new quoter

Everything else can go on either?

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -978,7 +935,7 @@ describe('alpha router integration', () => {
   });
 });
 
-describe('quote for other networks', () => {
+describe.only('quote for other networks', () => {
```
willpote, line 938: Oops

willpote (approved):

src/providers/on-chain-quote-provider.ts
zhongeric: V2Routes cannot be run on old Quoter though

snreynolds: Because we decided to just pass in one instance of `OnChainQuoteProvider` we now have to check at run time which quoter to pass the routes into. And I think the following should happen:
- We use the old quoter `QuoterV2` for all V3Routes (even though mixed route quoter will support V3Routes it will only support inputOnly routes)
- We use the new quoter `QuoterV3` (the mixed route quoter) for MixedRoutes and V2Routes that are exactInput (imo this second condition is still a little strange because the `QuoterV3` was written with the intention of just doing mixed routes, so we didn't consider exposing on chain quotes for v2 and thus, can now only do v2 exactInputs but I think its ok)

PR #159, opened 2022-08-12
Mixed routes improvements

zhongeric:

- **Other information**:


src/routers/alpha-router/alpha-router.ts
```diff
@@ -973,7 +974,7 @@ export class AlphaRouter
       /// and tradeType === EXACT_INPUT
       if (
         protocolsSet.has(Protocol.MIXED) &&
-        V2_SUPPORTED.includes(this.chainId) &&
+        (this.chainId === ChainId.MAINNET || this.chainId === ChainId.GÖRLI) &&
```
zhongeric, line 977: fyi i think the address for gorli is unfortunately different than the rest, the new address should be in the offboarding doc and that value should be set in the mapping in addresses.ts

snreynolds (approved):

willpote (approved):

PR #160, opened 2022-08-12
Add amountIn relative heuristic

zhongeric:

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1471,7 +1471,60 @@ export class AlphaRouter
     const V3poolsRaw = V3poolAccessor.getAllPools();
     const V2poolsRaw = V2poolAccessor.getAllPools();
 
-    const poolsRaw = [...V3poolsRaw, ...V2poolsRaw];
+    /// not just [Protocol.MIXED]
+    let chosenV2Pools: Pair[] = [];
+    if (
+      !routingConfig.forceMixedRoutes &&
+      routingConfig.protocols !== undefined &&
+      routingConfig.protocols.length > 1
+    ) {
+      /// Only consider V2 pools if amounts[100] (full amountIn) is of significant size compared to pool TVL
+      /// start out very conservative, let's do like 0.5%
+      const amountIn = amounts[amounts.length - 1]!;
+      /// TODO: direct swap pool is usually 10_000 (does not exist) so we could remove that
+      V2poolsRaw.forEach((pool) => {
+        if (
+          pool.involvesToken(amountIn.currency.wrapped) &&
+          amountIn.greaterThan(
+            pool.reserveOf(amountIn.currency.wrapped).divide(2000) // 0.005%
+          )
+        ) {
+          log.info(
+            {
+              pool: pool.toString(),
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1471,7 +1471,60 @@ export class AlphaRouter
     const V3poolsRaw = V3poolAccessor.getAllPools();
     const V2poolsRaw = V2poolAccessor.getAllPools();
 
-    const poolsRaw = [...V3poolsRaw, ...V2poolsRaw];
+    /// not just [Protocol.MIXED]
+    let chosenV2Pools: Pair[] = [];
+    if (
+      !routingConfig.forceMixedRoutes &&
```
snreynolds, line 1477: Why do we not want this logic when mixed routes are forced?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1471,7 +1471,60 @@ export class AlphaRouter
     const V3poolsRaw = V3poolAccessor.getAllPools();
     const V2poolsRaw = V2poolAccessor.getAllPools();
 
-    const poolsRaw = [...V3poolsRaw, ...V2poolsRaw];
+    /// not just [Protocol.MIXED]
```
snreynolds, line 1474: What is this comment? This only gets called in `getMixedRouteQuotes` so it should only be when we do specify we want mixed routes

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1471,7 +1471,60 @@ export class AlphaRouter
...
+          chosenV2Pools.push(pool);
+        }
+      });
+    }
+
+    log.info(
+      `Chose ${chosenV2Pools.length} out of ${V2poolsRaw.length} V2 pools for mixed routes`
+    );
+
+    const poolsRaw = [...V3poolsRaw, ...chosenV2Pools];
```
snreynolds, line 1527: `chosenV2Pools` is empty if the condition above is not met. what if `routingConfig.protocols.length = 1` and it's just mixed routes. We still want to be looking at v2 pools.

src/routers/alpha-router/alpha-router.ts
zhongeric: I think because if we are forcing mixedRoutes, we want to return ANY mixed route available, not necessarily only one that matches this heuristic criteria. I think the main use case for this that I envisioned was for the `protocols: [V3, V2, MIXED]` one and the `[]` case.

src/routers/alpha-router/alpha-router.ts
zhongeric: If we're only querying for `[Protocols.MIXED]` we would always want to return a mixed route if a route does exist. Using this heuristic, we only return mixed routes that we think are good. 

I guess there should be another conditional case where we use `...V2poolsRaw` instead of `...chosenV2Pools`

src/routers/alpha-router/alpha-router.ts
zhongeric: I think it meant that we should apply the heuristic below when the config is not just `Protocol.MIXED`

PR #163, opened 2022-08-17
feat: add network context to logs

marktoda:
This commit adds an optional context object to MetricsLogger which can be used
to add network id and network name to all logs. This may help alerting services
differentiate and bucket errors by network

Example log after this change
```
{"name":"Uniswap Smart Order Router",
"hostname":"toda-XPS-15-9510",
"pid":157780,
"chainId":1,
"networkName":"mainnet",
"level":30,
"key":"FindBestSwapRoute",
"value":13,
"unit":"Milliseconds",
"msg":"[Metric]: FindBestSwapRoute: 13 | Milliseconds",
"time":"2022-08-17T19:40:45.340Z",
"v":0}
```

### DIFF
--- a/cli/base-command.ts
+++ b/cli/base-command.ts
@@ -21,6 +21,7 @@ import {
   GasPrice,
   ID_TO_CHAIN_ID,
   ID_TO_PROVIDER,
+  ID_TO_NETWORK_NAME,
   IRouter,
   ISwapToRatio,
   ITokenProvider,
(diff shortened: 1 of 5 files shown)


willpote (approved):
FYI- the MetricLogger is only used in the SOR package, and mainly just exists as a shim/replacement for emitting metrics to a real metric service like sentry or AWS Cloudwatch. 

In the routing-API we implement IMetric such that it stores the [metrics in AWS Cloudwatch](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/router-entities/aws-metrics-logger.ts#L4-L13). 

For computing error rates per chain we actually need to emit completely new metrics for success and error by chain in the routing-api. We don't even have those base metrics being emitted yet, we just use the metrics which API Gateway emits automatically for every endpoint it runs.

Issue #58, opened 2022-01-31
Error: Convert JSBI instances to native numbers using `toNumber`.


fritzschoff:
- **I'm submitting a ...**
  [x] bug report
  [ ] feature request
  [ ] question about the decisions made in the repository
  [ ] question about how to use this project

- **Summary**

If you are using the smart-order-router and calling the `route` and for some random reasons, when the router calculates the gas price by converting the liquidity in the pool into a number, it throws an error. See error stack `V3HeuristicGasModelFactory.getHighestLiquidityUSDPool`
Pool address where this error ocurse is: `0x9957c4795ab663622db54fc48fda874da59150ff`

- **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)

``` 
Call Stack
JSBI.valueOf
node_modules/jsbi/dist/jsbi.mjs (1:1233)
baseGt
node_modules/@uniswap/smart-order-router/node_modules/lodash/lodash.js (3124:0)
baseExtremum
node_modules/@uniswap/smart-order-router/node_modules/lodash/lodash.js (2909:0)
Function.maxBy
node_modules/@uniswap/smart-order-router/node_modules/lodash/lodash.js (16407:0)
V3HeuristicGasModelFactory.getHighestLiquidityUSDPool
node_modules/@uniswap/smart-order-router/build/module/src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.js (201:25)
async V3HeuristicGasModelFactory.buildGasModel
node_modules/@uniswap/smart-order-router/build/module/src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.js (67:0)
async AlphaRouter.getV3Quotes
node_modules/@uniswap/smart-order-router/build/module/src/routers/alpha-router/alpha-router.js (370:0)
async AlphaRouter.route
node_modules/@uniswap/smart-order-router/build/module/src/routers/alpha-router/alpha-router.js (293:0)
```

willpote:
Hmm any way I can reproduce this locally? Can you share your tokenIn/tokenOut/amount and any other config you are using

fritzschoff:
follow this guide: https://docs.uniswap.org/sdk/guides/auto-router but with the pool address I provided

shreyaspapi:
I am getting the same error, has anyone got a solution?

hrishibhattu:
Same issue, any solutions?

[...]

PR #169, opened 2022-09-19
Add peer dependency to fix JSBI issue

hensha256:
Fixes the issues users are having in #58 - the issue occurs if they have v4 not v3 of JSBI. Given JSBI is not a direct dependency of this repo, I'm adding the version that we use as a peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,9 @@
     "typechain": "^5.0.0",
     "typescript": "^4.2.2"
   },
+  "peerDependencies": {
+    "jsbi": "3.2.5"
+  },
   "files": [
     "build/main",
     "build/module",


package.json
```diff
@@ -96,6 +96,9 @@
     "typechain": "^5.0.0",
     "typescript": "^4.2.2"
   },
+  "peerDependencies": {
+    "jsbi": "3.2.5"
```
willpote, line 100: Maybe this should be "^3.2.5", otherwise we are requiring the peer has exactly 3.2.5 which is probably too strict/restrictive?

willpote (approved):
Feel free to update if you agree with the comment and merge without another approval

kevjin:
> Given JSBI is not a direct dependency of this repo, I'm adding the version that we use as a peer dependency.

Think it is a direct dep, as can be seen here, https://github.com/Uniswap/smart-order-router/blob/ff7af78a78a8df0328d8cbe2f0484493030952c5/src/util/amounts.ts#L7
Having JSBI as a peerdep appears to be causing some issues for our bundler, rollup, and wonder if y'all can change this back to a direct dep? thanks! @hensha256 @willpote 
i added a tracking issue here: https://github.com/Uniswap/smart-order-router/issues/191

PR #175, opened 2022-10-12
Added param for page size for v2 subgraph provider

raghava-pamula:
- **What kind of change does this PR introduce?** Feature to set pageSize for v2SubgraphProvider

- **What is the current behavior?** Uses pageSize = 1000

- **What is the new behavior (if this is a feature change)?**. Uses whatever pageSize is passed, 1000 by default

- **Other information**: This is to try to work around this issue: https://github.com/Uniswap/v2-subgraph/issues/78 

By reducing page size we may be able to get the subgraph query to work more reliably


willpote (commented):
package lock shouldn't change for this

willpote (approved):

PR #183, opened 2022-11-02
Raghava/add check for sufficient user balance before simulating tx

raghava-pamula:
Adds check before simulating transactions from addresses that don't have sufficient balance. This prevents us from simulating when we already know the transaction would fail on chain.

It simulates no matter what if `fromAddress` field is set
https://uniswaplabs.atlassian.net/browse/TRD-159

Simulates as normal if user has sufficient token balance, otherwise returns flag `simulationAttempted = false`

- **Other information**:
Will also help with integrating tenderly with web for gas estimates


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1885,6 +1900,31 @@ export class AlphaRouter
...
+      } else {
+        const tokenContract = Erc20__factory.connect(
+          neededBalance.currency.address,
+          this.provider
+        );
+        balance = await tokenContract.balanceOf(fromAddress);
+      }
+      return balance.gte(BigNumber.from(neededBalance.quotient.toString()));
+    } catch (e) {
+      log.info(e, 'Error while checking user balance');
```
rileydcampbell, line 1923:
```suggestion
      log.error(e, 'Error while checking user balance');
```
Should we log this as an error instead?

rileydcampbell (commented):
lgtm, just one nit.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -1140,87 +1098,28 @@ describe('alpha router integration', () => {
                 recipient: alice._address,
                 slippageTolerance: SLIPPAGE,
                 deadline: parseDeadline(360),
-                simulate: { fromAddress: WHALES(tokenIn) },
+                simulate: { fromAddress: '0xeaf1c41339f7D33A2c47f82F7b9309B5cBC83B5F' },
```
ConjunctiveNormalForm, line 1101: this change looks rather arbitrary

ConjunctiveNormalForm (commented):
I'm not sure why we are removing some of the unit tests

test/integ/routers/alpha-router/alpha-router.integration.test.ts
raghava-pamula: this test is checking to make sure simulation is not called if the user does not have sufficient token balance. The test will fail without this change because the old address had sufficient balance

raghava-pamula:
> I'm not sure why we are removing some of the unit tests

That test checks to see if the FallbackTenderlySimulator is getting called when appropriate. I removed it because I thought it was being checked in the integ-test here: https://github.com/Uniswap/smart-order-router/pull/183/files#diff-b1a09e5a7156f19225eba6ca4e2e78d0942580ff1118bab472bf0ea3b1294904R908

Trying to add the unit test back now, I need to mock the `userHasSufficientBalance` method for it to work, it was failing before I removed it.

ConjunctiveNormalForm (approved):

rileydcampbell (approved):

raghava-pamula:
There was one additional case which I now addressed. For ETH -> ERC20 exact-out case, the quote is returned in WETH even though user will pay in ETH (I think this is because we send the users ETH to the WETH contract before we do anything else, and then we send the WETH to the SwapRouter contract).

I handled this case and added test coverage for it in the integ-tests.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1109,23 +1111,38 @@ export class AlphaRouter
...
+      if (
+        await this.userHasSufficientBalance(
+          fromAddress,
+          tradeType,
+          amount,
+          // Quote will be in WETH even if quoteCurrency is ETH
+          // So we init a new CurrencyAmount object here
+          CurrencyAmount.fromRawAmount(quoteCurrency, quote.quotient.toString())
+        )
+      ) {
```
willpote, line 1124: Why not make this part of the simulator class itself? As written we make an assumption that no simulator can ever succeed unless the from address has balance, but I may implement an `ISimulator` which can actually handle this case (maybe i transfer tokens to the from address on a fork, or something like that)

src/routers/router.ts
```diff
@@ -84,6 +84,10 @@ export type SwapRoute = {
    * Flag that is true if and only if simulation is requested and simulation fails
    */
   simulationError?: boolean;
+  /**
+   * Flag that is true if and only if simulation is requested and also attempted
+   */
+  simulationAttempted?: boolean;
```
willpote, line 90: What do we need this flag for? Can't the user not having a balance just be part of simulationError? We should try and avoid adding optional flags like this to the returned object- it complicates logic for clients

raghava-pamula: removed it

src/routers/alpha-router/alpha-router.ts
raghava-pamula: moved it to the simulator. Let me know if i should make it a static method, i was not sure.

raghava-pamula: Also, I changed ISimulator to an abstract class to handle this

src/routers/router.ts
raghava-pamula: [i made it so that if the swapParams are not set, then it will also set simulationError bit to true](https://github.com/Uniswap/smart-order-router/pull/183/files#diff-2fe4a092f36b8091262f7035638dc00530816382f3a1cca86927a6fd712e9c04R1137)

This will make it easy for web integration. Otherwise, web could request a quote without fromAddress/swapParams set, and they would receive a quote response without simulationError bit set. Based on the logic of only having simulationError and not simulationAttempted, they would think we successfully simulated the gas estimate for it. But that wouldn't be the case.

raghava-pamula:
also, added spec file for simulation

test/unit/routers/alpha-router/alpha-router.test.ts
```diff
@@ -1850,44 +1944,39 @@ describe.only('alpha router', () => {
       sinon.assert.notCalled(mockOnChainQuoteProvider.getQuotesManyExactOut);
     });
 
-    test('succeeds to route and generates calldata on v3 only and simulates', async () => {
+    test('succeeds to route and generates calldata on v2 only', async () => {
```
willpote, line 1947: Whys this and the below unit tests changed?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1125,7 +1132,9 @@ export class AlphaRouter
       return swapRouteWithSimulation;
     }
 
-    return swapRoute;
+    // anytime we don't have a simulated gasLimit estimate
+    // we set simulationError to true
+    return { ...swapRoute, simulationError: true };
```
willpote, line 1137: I don't think this is right, I think we should just leave it undefined if we didn't actually run a simulation

raghava-pamula: sure

test/unit/routers/alpha-router/alpha-router.test.ts
raghava-pamula: I changed the order of the specs

raghava-pamula: I put the simulation specs at the end of each describe to organize better. Also I removed some unnecessary params from the simulation specs.

raghava-pamula: also, i added more specs in simulation-provider.test.ts, so test coverage actually increased.

raghava-pamula: i can change the order of the tests back if you think it should not change, just lmk

willpote (approved):

PR #186, opened 2022-11-18
Raghava/add simulation status enum

raghava-pamula:
Feature

returns simulationError = true if simulation failed or insufficient funds

returns simulationStatus enum which distinguishes between simulation failed and insufficient funds

- **Other information**:
followup PR needed in routing-api to stop alerting for insufficient funds


ConjunctiveNormalForm (commented):
will anything break externally, given we are removing `simulationError` field?

raghava-pamula:
> will anything break externally, given we are removing `simulationError` field?

should not, currently only routing-api uses this field

ConjunctiveNormalForm (approved):

willpote (approved):
We need to be careful how we roll this out to the API as it is a backward incompatible change to the API response

PR #187, opened 2022-11-30
Fix exception when gasToken has less decimals

coopermaruyama:
Addresses aforementioned todo

Bugfix

- **What is the current behavior?**
Getting a quote on goerli throws with 'Exponent must be positive'

Works now

### DIFF
--- a/src/routers/alpha-router/functions/best-swap-route.ts
+++ b/src/routers/alpha-router/functions/best-swap-route.ts
@@ -467,7 +467,6 @@ export async function getBestSwapRouteBy(
   // For each gas estimate, normalize decimals to that of the chosen usd token.
   const estimatedGasUsedUSDs = _(bestSwap)
     .map((routeWithValidQuote) => {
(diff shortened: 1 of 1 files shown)


coopermaruyama:
still relevant.. there's literally a TODO right in the comments to eventually fix this

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -467,7 +467,6 @@ export async function getBestSwapRouteBy(
   // For each gas estimate, normalize decimals to that of the chosen usd token.
   const estimatedGasUsedUSDs = _(bestSwap)
     .map((routeWithValidQuote) => {
-      // TODO: will error if gasToken has decimals greater than usdToken
```
jsy1218, line 470: We discussed this is a change of behavior, and would like to get feedback on the original author(s) (cc @snreynolds @zhongeric ) whether we'd like to pursue this path. The comment says erroring it out in such case, and we'd want to understand better the reasoning behind it.

PR #189, opened 2022-11-30
Add eth estimate gas simulator

raghava-pamula:
Adds a new Simulator for gas estimation, to be used in web gas estimates

EthEstimateGasSimulator estimates gas limit and returns a new swap route

- **Other information**:
Follow up PR needed in Routing-API to use the new simulator


raghava-pamula:
almost done writing specs for this change

raghava-pamula:
added specs

src/providers/eth-estimate-gas-provider.ts
```diff
@@ -0,0 +1,127 @@
...
+      );
+
+      estimatedGasUsed = await this.provider.estimateGas({
+        data: route.methodParameters!.calldata,
+        to: route.methodParameters!.to,
+        from: fromAddress,
+        value: BigNumber.from(
+          currencyIn.isNative ? route.methodParameters!.value : '0'
+        ),
+      });
```
willpote, line 64: Shouldn't try/catch this and return `simulationStatus: SimulationStatus.Failed` here if this errors out? i.e. We attempted to do a simulation and something went wrong

src/providers/eth-estimate-gas-provider.ts
```diff
@@ -0,0 +1,127 @@
...
+    } else if (swapOptions.type == SwapType.SWAP_ROUTER_02) {
+      log.info(
+        { methodParameters: route.methodParameters },
+        'Simulating using eth_estimateGas on SwapRouter02'
+      );
+
+      estimatedGasUsed = await this.provider.estimateGas({
+        data: route.methodParameters!.calldata,
+        to: route.methodParameters!.to,
```
willpote, line 59: You just copy pasted all this logic from within the TenderlySimulationProvider. 

We should follow the DRY principle, and refactor the TenderlySimulationProvider so that it shares the logic.

src/providers/eth-estimate-gas-provider.ts
```diff
@@ -0,0 +1,127 @@
...
+        from: fromAddress,
+        value: BigNumber.from(
+          currencyIn.isNative ? route.methodParameters!.value : '0'
+        ),
+      });
+    } else {
+      throw new Error(`Unsupported swap type ${swapOptions}`);
+    }
+
+    estimatedGasUsed = estimatedGasUsed.mul(BigNumber.from(1.2));
```
willpote, line 69: Extract 1.2 to a constant. Also below we did `1.2`, here we do `BigNumber.from(1.2)`

test/unit/providers/simulation-provider.test.ts
```diff
@@ -99,13 +97,99 @@ describe('fallback tenderly simulator', () => {
...
+        blockNumber: blockNumber,
+        simulationStatus: SimulationStatus.Succeeded,
+    };
+
+    beforeEach(() => {
+        simulateTxStub = sinon.stub(simulator, <any>'simulateTransaction');
+        simulateTxStub.resolves(swaproute);
+        ethEstimateGasStub = sinon.stub(simulator, <any>'ethEstimateGas');
+        ethEstimateGasStub.resolves(swaproute)
+    });
```
willpote, line 134: These unit tests don't really make sense- you've mocked out the logic from the class that we are testing.. so what logic is actually being tested at all?

test/unit/providers/simulation-provider.test.ts
```diff
@@ -99,13 +97,99 @@ describe('fallback tenderly simulator', () => {
...
+    
+    afterEach(() => {
+        sinon.restore();
+    });
+    test('simulates when user has sufficient balance and token is approved', async () => {
+        sinon.stub(simulator, <any>'userHasSufficientBalance').resolves(true);
+        sinon.stub(
+            simulator,
+            <any>'checkTokenApproved',
+        ).resolves(true);
```
willpote, line 144: Since you've mocked out the function that calls this, this does nothing. Change .resolves to 'false' and the test will still pass.

willpote (changes requested):
No integration tests?

src/providers/eth-estimate-gas-provider.ts
raghava-pamula: okay, i'll add try/catch. I think I remember in the past we decided to not try/catch because we expect provider.estimateGas to work. I think it makes sense to try/catch now.

src/providers/eth-estimate-gas-provider.ts
raghava-pamula: Okay, extracting it out into util file and call the same method in both places.

willpote: You could make TenderlySimulationProvider depend on EthEstimateProvider, so pass in EthEstimateProvider in the constructor

test/unit/providers/simulation-provider.test.ts
raghava-pamula: fixed the tests

src/providers/eth-estimate-gas-provider.ts
raghava-pamula: yeah, that's what I'm doing

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -2782,9 +2910,9 @@ describe('quote for other networks', () => {
                     .equalTo(swap.estimatedGasUsedQuoteToken)
                 );
 
-                // Expect tenderly simulation to be successful
-                expect(swap.simulationStatus).toEqual(
-                  SimulationStatus.Succeeded
+                // Expect Eth Estimate Gas to not fail
+                expect(swap.simulationStatus).not.toEqual(
+                  SimulationStatus.Failed
```
willpote, line 2915: Why are we not checking success anymore? These are the simulation tests, so they must succeed.

src/providers/simulation-provider.ts
```diff
@@ -20,6 +20,8 @@ export enum SimulationStatus {
   Unattempted = 0,
   Failed = 1,
   Succeeded = 2,
+  InsufficientBalance = 3,
+  NotApproved = 4,
```
willpote, line 24: When is unattempted used now? Should we remove?

test/unit/providers/simulation-provider.test.ts
```diff
@@ -4,108 +4,206 @@ import { Percent } from '@uniswap/sdk-core';
...
+        const swapRoute = await simulator.simulate(
+            fromAddress,
+            swapOptions,
+            swaproute,
+            amount,
+            quote
+        );
+        expect(ethEstimateGasStub.calledOnce).toBeFalsy();
```
willpote, line 186: `.called`

test/unit/providers/simulation-provider.test.ts
```diff
@@ -4,108 +4,206 @@ import { Percent } from '@uniswap/sdk-core';
...
+        trade: trade as Trade<any, any, any>,
+        route: route,
+        blockNumber: blockNumber,
+        simulationStatus: SimulationStatus.Succeeded,
+    };
+
+    beforeEach(() => {
+        ethEstimateGasStub = sinon.stub(simulator, <any>'ethEstimateGas');
+        ethEstimateGasStub.resolves(swaproute)
+    });
```
willpote, line 149: In these unit tests you are mocking functions of the class, so the result is we aren't testing all the logic in the class. i.e. all the logic in `ethEstimateGas` about `swap_router_02` vs `universal_router`, gas multiplying, setting status to failed or not, is all untested.

The correct way to do it is to mock the dependencies that are injected into the class. So you should be mocking the jsonRpcProvider to return different things

src/providers/eth-estimate-gas-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+        'Simulating using eth_estimateGas on Universal Router'
+      );
+      estimatedGasUsed = await this.provider.estimateGas({
+        data: route.methodParameters!.calldata,
+        to: route.methodParameters!.to,
+        from: fromAddress,
+        value: BigNumber.from(
+          currencyIn.isNative ? route.methodParameters!.value : '0'
+        ),
+      });
```
willpote, line 50: Should be in a try/catch as well like below? 

See below comment about mocking functions on the class under test - this is the sort of thing that can be missed.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -2062,11 +2090,113 @@ describe('alpha router integration', () => {
...
+              amount,
+              getQuoteToken(tokenIn, tokenOut, tradeType),
+              tradeType,
+              {
+                type: SwapType.SWAP_ROUTER_02,
+                recipient: alice._address,
+                slippageTolerance: SLIPPAGE,
+                deadline: parseDeadline(360),
+                simulate: { fromAddress: WHALES(tokenIn) },
+              },
```
willpote, line 2175: Lets add a test with Universal Router, Swap Router 02 will be deprecated soon

willpote (changes requested):

test/unit/providers/simulation-provider.test.ts
raghava-pamula: I see, doing that now. And will keep in mind for the future.

test/unit/providers/simulation-provider.test.ts
raghava-pamula: fixed

test/integ/routers/alpha-router/alpha-router.integration.test.ts
raghava-pamula: done

src/providers/simulation-provider.ts
raghava-pamula: Removed it and replaced with NotSupported (for cases like Celo)

willpote (approved):
1 nit

PR #195, opened 2023-01-20
Fix quoteAdjustedForGas values for low liq pools

zhongeric:
We want to get the gas cost in terms of the quote token.
        Sometimes, the quote token / native token pair is very illiquid and the price is incorrect.
        We also have access to the input token / native token pair, which is more likely to be liquid.
          - if it's not, then it is an illiquid direct swap trade
        We can get the gas cost in terms of the input token, 
          using the inputTokenNativePool and quoting for the amount of ETH required for gas
        Then, we can use the execution price to get the respective amounts of quote tokens
        Then, we can take the minimum of the initial quoted gas cost in terms of quote token
          and our calculated one.

Original failing swap: USDC (1000) -> USDP
```
./bin/cli quote --tokenIn 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 --tokenOut 0x8E870D67F660D95d5be530380D0eC0bd388289E1 --amount 1000 --exactIn --recipient 0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B --protocols v3
```
Response
```
{
  nativeQuoteTokenPrice: '340257000000000000000000000000000000000',
  nativeInputTokenPrice: '1582.44',
  gasCostInTermsOfQuoteToken: '1032483198235071252566720567390277190.582140643337787585',
  gasCostInTermsOfInputToken: '4.801802',
  executionPrice: '1.00039',
  syntheticGasCostInTermsOfQuoteToken: '4.80369'
}
Best Route:
[V3] 100.00% = USDC -- 0.05% [0x5720EB958685dEEEB5AA0b34F677861cE3a8c7F5] --> USDP
	Raw Quote Exact In:
		1000.39
	Gas Adjusted Quote In:
		995.58

Gas Used Quote Token: 4.803690
Gas Used USD: 4.817489
```
- `syntheticGasCostInTermsOfQuoteToken` is calculated by first getting the gasCostInTermsOfInputToken, then doing `executionPrice.quote(gasCostInTermsOfInputToken)` to get it in terms of QuoteToken.
- Note how we use the `syntheticGasCostInTermsOfQuoteToken` over the `gasCostInTermsOfQuoteToken`

Example ERC20 -> ERC20: UNI (1000) -> SOCKS
Response:
```
{
  nativeQuoteTokenPrice: '0.0476681',
  nativeInputTokenPrice: '251.091',
  gasCostInTermsOfQuoteToken: '0.000506562142531005',
  gasCostInTermsOfInputToken: '2.668312573375472718',
  executionPrice: '0.000106902',
  syntheticGasCostInTermsOfQuoteToken: '0.000285246'
}

Best Route:
[V3] 100.00% = UNI -- 0.3% [0x1d42064Fc4Beb5F8aAF85F4617AE8b3b5B8Bd801] --> WETH -- 1% [0xc75a99Fa00803896349891F8471Da3614bd07564] --> SOCKS
	Raw Quote Exact In:
		0.18
	Gas Adjusted Quote In:
		0.18

Gas Used Quote Token: 0.000299
Gas Used USD: 10.268416
```


src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -259,6 +269,81 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
...
+        Sometimes, the quote token / native token pair is very illiquid and the price is incorrect.
+        We also have access to the input token / native token pair, which is more likely to be liquid.
+          - if it's not, then it is an illiquid direct swap trade
+        We can get the gas cost in terms of the input token, 
+          using the inputTokenNativePool and quoting for the amount of ETH required for gas
+        Then, we can use the execution price to get the respective amounts of quote tokens
+        Then, we can take the minimum of the initial quoted gas cost in terms of quote token
+          and our calculated one.
+      */
+      try {
```
willpote, line 283: Why surround with a try? I don't see any reason why this should fail

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -259,6 +269,81 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
...
+            gasCostInTermsOfInputToken
+          );
+
+          // If the cost in quote tokens over the execution price is lessThan the cost found using the quoteToken/native pool
+          // then we use our synthetic gas cost in tokens
+          if (
+            syntheticGasCostInTermsOfQuoteToken.lessThan(
+              gasCostInTermsOfQuoteToken.asFraction
+            )
+          ) {
```
willpote, line 316: Thinking about this more, I'm not sure that if the synthetic is smaller then it is *always* better. Like it would be possible for there to be a borked amount/native pool with a price such that it always returns a smaller + less accurate value. It just happens that the USDP pool is broken in a way where the quoting ETH returns a huge value.

Still though, I think its better to prefer the smaller value since big values lead to crazy results because the adjusted quotes go negative. Where as with smaller values we kind of just fallback to ignoring gas when picking the best route- which feels a lot better.

Does that makes sense? If so i think we should call it out in the comment

willpote (approved):
Nice! Mostly nits!

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -65,7 +65,8 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
     chainId,
     gasPriceWei,
     v3poolProvider: poolProvider,
-    token,
+    inputToken,
+    quoteToken,
```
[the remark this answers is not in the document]
zhongeric: Yeah, makes sense. Going to use amountToken

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -259,6 +269,81 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
         throw err;
       }
 
+      /*
+        We want to get the gas cost in terms of the quote token.
+        Sometimes, the quote token / native token pair is very illiquid and the price is incorrect.
+        We also have access to the input token / native token pair, which is more likely to be liquid.
```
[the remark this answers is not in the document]
zhongeric: hmm yeah you're right, I was thinking that it is likely that at least one of the native pools is liquid, otherwise it would be a direct swap between two illiquid coins and we would fail to give a good quote anyways.

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
zhongeric: Was mirroring the try/catch on line 255:
```

      let gasCostInTermsOfQuoteToken: CurrencyAmount;
      try {
        // native token is base currency
        gasCostInTermsOfQuoteToken = nativeTokenPrice.quote(
          totalGasCostNativeCurrency
        ) as CurrencyAmount;
      } catch (err) {
        log.info(
          {
            nativeTokenPriceBase: nativeTokenPrice.baseCurrency,
            nativeTokenPriceQuote: nativeTokenPrice.quoteCurrency,
            gasCostInEth: totalGasCostNativeCurrency.currency,
          },
          'Debug eth price token issue'
        );
        throw err;
      }
```

zhongeric: yeah not sure its needed, unless `<Price>.quote()` can fail

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
zhongeric: Yeah I agree. If the synthetic price calculated is extremely small due to error, we will basically just return the quote w/o gas consideration, and at least the user receives a quote - we use the gasEstimate in gwei anyways when submitting so shouldnt be a huge issue? Maybe it would hit slippage, i'm not sure as I haven't looked at that code in a while.

zhongeric: Additionally, something interesting that I've noticed is that the vast majority of the integ tests and cli quotes that I've run result in a smaller syntheticGasCost... it's smaller by a small amount, and I'm not quite sure why.

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -225,11 +235,11 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
...
-          `Unable to find ${nativeCurrency.symbol} pool with the quote token, ${token.symbol} to produce gas adjusted costs. Route will not account for gas.`
+          `Unable to find ${nativeCurrency.symbol} pool with the quote token, ${quoteToken.symbol} to produce gas adjusted costs. Route will not account for gas.`
         );
         return {
           gasEstimate: baseGasUse,
-          gasCostInToken: CurrencyAmount.fromRawAmount(token, 0),
+          gasCostInToken: CurrencyAmount.fromRawAmount(quoteToken, 0),
           gasCostInUSD: CurrencyAmount.fromRawAmount(usdToken, 0),
         };
       }
```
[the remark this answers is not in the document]
zhongeric: Sounds good, that's a good addition

willpote (approved):

PR #199, opened 2023-02-24
Refactoring Alpha Router route method

mikeki:
- **What kind of change does this PR introduce?** Refactor of code

- **What is the current behavior?** Boolean logic is convoluted and some logic could be pulled into their own helper methods.

- **What is the new behavior (if this is a feature change)?** Boolean logic has been simplified and helper methods have been created.

- **Other information**: No behavior should have changed, this is purely a refactor.


src/routers/alpha-router/alpha-router.ts
```diff
@@ -877,34 +866,23 @@ export class AlphaRouter
...
+    const v2ProtocolSpecified = protocolsSet.has(Protocol.V2)
+    const mixedProtocolSpecified = protocolsSet.has(Protocol.MIXED)
+    const v2SupportedInChain = V2_SUPPORTED.includes(this.chainId)
+    const tradeTypeIsExactInput = tradeType === TradeType.EXACT_INPUT
+    const isMainnet = this.chainId === ChainId.MAINNET
+    const isGorli = this.chainId === ChainId.GÖRLI
+
+    // Maybe Quote V3 - if V3 is specified, or no protocol is specified
+    if (v3ProtocolSpecified || noProtocolsSpecified) {
+      // Open Question: Should we quote V3 if only V2 is specified but v2 is not supported?
```
kmairin, line 884: IMHO, if the request wants a v2 quote and v2 is not supported we should return a 4xx? so the client knows that v2 is not supported for this protocol

mikeki: Good call, I think that indirectly happens here: https://github.com/Uniswap/smart-order-router/pull/199/files/040d5685188d9e1d93cc976f72a792947467c4dd#diff-2fe4a092f36b8091262f7035638dc00530816382f3a1cca86927a6fd712e9c04L1033

The boolean logic should in theory be the same as it was before, but easier to understand and simplified

src/routers/alpha-router/alpha-router.ts
```diff
@@ -877,34 +866,23 @@ export class AlphaRouter
...
+      log.info({ protocols, swapType: tradeType }, 'Routing across V3');
```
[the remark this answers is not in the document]
zhongeric: Historically these have been used for debugging, iirc if they are not added to the `metric: IMetric` they are not logged in analytics

src/routers/alpha-router/alpha-router.ts
```diff
@@ -929,92 +912,34 @@ export class AlphaRouter
...
-        );
-      }
+    }
+
+    // Maybe Quote mixed routes
+    // if MixedProtocol is specified or no protocol is specified and v2 is supported AND tradeType is ExactIn
+    // AND is Mainnet or Gorli
+    const shouldQueryMixedProtocol = mixedProtocolSpecified || (noProtocolsSpecified && v2SupportedInChain)
+    if (shouldQueryMixedProtocol && tradeTypeIsExactInput && (isMainnet || isGorli)) {
+      log.info(
```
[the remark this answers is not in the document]
mikeki: It would be great if we added a linting/code prettifier task to this repo, my VSCode formatted this code and made some unwanted changes, and I'm not sure how to disable the auto-formatting locally 😅

src/routers/alpha-router/alpha-router.ts
zhongeric: Yes we should return no route found in this case, however routing-api should still return 200. I don't think we ever return 400 unless malformed input?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -877,34 +866,23 @@ export class AlphaRouter
...
 
-    if (
-      (protocolsSet.size == 0 ||
-        (protocolsSet.has(Protocol.V2) && protocolsSet.has(Protocol.V3))) &&
-      V2_SUPPORTED.includes(this.chainId)
-    ) {
-      log.info({ protocols, tradeType }, 'Routing across all protocols');
+    const [v3gasModel, mixedRouteGasModel] = await this.getGasModels(gasPriceWei, amount.currency.wrapped, quoteToken, metric);
+
+    const noProtocolsSpecified = protocolsSet.size == 0
```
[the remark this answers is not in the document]
mikeki: gotcha! so the logging should use different logic than the selection of protocols

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
mikeki: agree, for some reason thats the log we had before and I just kept it that way since i'm not sure if somebody is using it with that name

mikeki:
Replying to comments

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1142,6 +1067,68 @@ export class AlphaRouter
...
+      return {
+        currencyIn: quoteCurrency,
+        currencyOut: amount.currency
+      }
+    }
+  }
+
+  private async getGasPriceWei(metric: IMetric): Promise<BigNumber> {
+    // Get an estimate of the gas price to use when estimating gas cost of different routes.
+    const beforeGas = Date.now();
```
kmairin, line 1086: maybe `beforeGasDate`?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -929,92 +912,34 @@ export class AlphaRouter
...
+          mixedRouteGasModel,
+          tradeType,
+          routingConfig
+        )
+      );
     }
 
     const routesWithValidQuotesByProtocol = await Promise.all(quotePromises);
 
+    // Manual reduce of Array<RouteWithValidQuote[],CandidatePoolsBySelectionCriteria>
```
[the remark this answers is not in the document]
mikeki: Oh the comment is just explaining whats happening, because the code was a bit confusing, but i think we need to manually reduce because we are pulling 2 kinds of arrays instead of just 1

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
zhongeric: Yea, I dont think anyone is using it because log.info is only exposed in debug logs and not tracked anywhere?

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
zhongeric: wdm by "the logging"?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -843,12 +842,10 @@ export class AlphaRouter
       { blockNumber }
     );
 
-    const { protocols } = routingConfig;
+    const protocols = routingConfig.protocols ?? [];
```
snreynolds, line 845: We can just create a set right away here
`const protocolsSet = new Set(routingConfig.protocols ??  []);` and just log the protocolsSet

src/routers/alpha-router/alpha-router.ts
```diff
@@ -929,92 +911,33 @@ export class AlphaRouter
...
-            routingConfig
-          )
-        );
-      }
+    }
+
+    // Maybe Quote mixed routes
+    // if MixedProtocol is specified or no protocol is specified and v2 is supported AND tradeType is ExactIn
+    // AND is Mainnet or Gorli
+    const shouldQueryMixedProtocol = mixedProtocolSpecified || (noProtocolsSpecified && v2SupportedInChain)
```
snreynolds, line 919: hm, lots of conditionals defined above that I think we could probably combine/condense? Maybe something like
- replacing the variables `isMainnet`, `isGorli`, and `tradeTypeIsExactInput` with a variable like `mixedProtocolsAllowed` as we only use them to check for mixed routes
- so then `mixedProtocolsAllowed` could be 
```
const mixedProtocolsAllowed = tradeType === TradeType.EXACT_INPUT && (this.chainId === ChainId.MAINNET || this.chainId === ChainId.GÖRLI)
```
- then changing the if statement to

```
if (mixedProtocolsSpecifed || (noProtocolsSpecifed && v2SupportedInChain) && mixedProtocolsAllowed) {...}
```

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1142,6 +1065,69 @@ export class AlphaRouter
...
+    metric.putMetric(
+      'GasPriceLoad',
+      Date.now() - beforeGasTimestamp,
+      MetricLoggerUnit.Milliseconds
+    );
+
+    return gasPriceWei;
+  }
+
+  private async getGasModels(gasPriceWei: BigNumber, amountToken: Token, quoteToken: Token, metric: IMetric): Promise<[IGasModel<V3RouteWithValidQuote>, IGasModel<MixedRouteWithValidQuote>]> {
```
snreynolds, line 1098: same comment here, but I don't think you need to pass in metric

snreynolds (changes requested):

src/routers/alpha-router/alpha-router.ts
mikeki: I thought so too, but we log the `protocols` as an Array, logging it as a set would change the structure of the log.

Actually, do you know if there's any particular reason of why we create a set? We could accomplish the same thing by doing `.includes` on the Array instead of `.has` on the Set

kmairin (approved):
thanks for addressing the comments!

src/routers/alpha-router/alpha-router.ts
```diff
@@ -929,92 +882,32 @@ export class AlphaRouter
...
+          mixedRouteGasModel,
+          tradeType,
+          routingConfig
+        )
+      );
     }
 
     const routesWithValidQuotesByProtocol = await Promise.all(quotePromises);
 
+    // The following block of codoe performs a manual reduce operation
```
willpote, line 908: Nit: code

willpote (approved):
Nice refactor!

src/routers/alpha-router/alpha-router.ts
zhongeric: I think just to auto filter out any accidental duplicates?

mikeki: yup, I ended up applying some logic to get the best of both worlds

PR #200, opened 2023-02-28
Fix pool resolution for gas + routeString, cli fix

zhongeric:

- **Other information**:

### DIFF
--- a/src/util/gas-factory-helpers.ts
+++ b/src/util/gas-factory-helpers.ts
@@ -41,7 +41,7 @@ export async function getV2NativePool(
         reserve0: pool?.reserve0.toExact(),
         reserve1: pool?.reserve1.toExact(),
       },
-      `Could not find a valid WETH pool with ${token.symbol} for computing gas costs.`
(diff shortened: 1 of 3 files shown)


src/util/gas-factory-helpers.ts
```diff
@@ -56,15 +56,25 @@ export async function getHighestLiquidityV3NativePool(
 ): Promise<Pool | null> {
   const nativeCurrency = WRAPPED_NATIVE_CURRENCY[token.chainId as ChainId]!;
 
-  const nativePools = _([FeeAmount.HIGH, FeeAmount.MEDIUM, FeeAmount.LOW])
+  const nativePools = _([
+    FeeAmount.HIGH,
+    FeeAmount.MEDIUM,
+    FeeAmount.LOW,
+    FeeAmount.LOWEST,
```
zhongeric, line 63: Some chains may not have `FeeAmount.LOWEST`, however it should be fine as they will be filtered out

willpote (approved):

PR #205, opened 2023-03-09
feat: add support for Optimism Goerli

smartcontracts:
Adds support for Optimism Goerli in the various different places where support is required.

---

Welp, I also ran `npm run fix` which linted a bunch of things, so a lot of this diff is purely linter results.


src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -95,7 +95,11 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
       };
       let l1Used = BigNumber.from(0);
       let l1FeeInWei = BigNumber.from(0);
-      if (chainId == ChainId.OPTIMISM || chainId == ChainId.OPTIMISTIC_KOVAN) {
+      if (
+        chainId == ChainId.OPTIMISM ||
+        chainId == ChainId.OPTIMISTIC_KOVAN ||
```
zhongeric, line 100: Is kovan still used? We could deprecate

PR #209, opened 2023-03-17
Allow more granularity when selection topN for second hops

mikeki:
feature
Currently we indiscriminately select the topN for all second hops
This feature allows us to specify more granularity for second hops given a token address
- **Other information**:

### DIFF
--- a/cli/commands/quote.ts
+++ b/cli/commands/quote.ts
@@ -5,13 +5,7 @@ import { Currency, Percent, TradeType } from '@uniswap/sdk-core';
 import dotenv from 'dotenv';
 import _ from 'lodash';
 
-import {
-  ID_TO_CHAIN_ID,
-  nativeOnChain,
-  parseAmount,
-  SwapRoute,
-  SwapType,
-} from '../../src';
+import { ID_TO_CHAIN_ID, nativeOnChain, parseAmount, SwapRoute, SwapType, } from '../../src';
 import { NATIVE_NAMES_BY_ID, TO_PROTOCOL } from '../../src/util';
 import { BaseCommand } from '../base-command';
 
@@ -55,6 +49,7 @@ export class Quote extends BaseCommand {
       topN,
       topNTokenInOut,
       topNSecondHop,
+      topNSecondHopForTokenAddressRaw,
       topNWithEachBaseToken,
       topNWithBaseToken,
       topNWithBaseTokenInSet,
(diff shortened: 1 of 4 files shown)


cli/commands/quote.ts
```diff
@@ -87,26 +101,20 @@ export class Quote extends BaseCommand {
       }
     }
 
-    const chainId = ID_TO_CHAIN_ID(chainIdNumb);
-
-    const log = this.logger;
-    const tokenProvider = this.tokenProvider;
-    const router = this.router;
```
zhongeric, line 94: These unused?

mikeki: I moved them up in the definition because I wanted to access the log before it was defined, let me revert this change because it's not necessary

cli/commands/quote.ts
```diff
@@ -70,6 +65,25 @@ export class Quote extends BaseCommand {
...
+
+    const topNSecondHopForTokenAddress = new Map();
+    topNSecondHopForTokenAddressRaw.split(',').forEach((entry) => {
+      if (entry != '') {
+        const entryParts = entry.split('|');
+        if (entryParts.length != 2) {
+          throw new Error(
+            'flag --topNSecondHopForTokenAddressRaw must be in format tokenAddress|topN,...');
+        }
+        const topNForTokenAddress: number = +entryParts[1]!;
```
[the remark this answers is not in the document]
mikeki: good suggestion, thank you.

cli/commands/quote.ts
zhongeric: oh i see, missed it up there

src/routers/alpha-router/alpha-router.ts
```diff
@@ -220,9 +220,18 @@ export type ProtocolPoolSelection = {
...
   topNSecondHop: number;
+  /**
+   * Given the topNTokenInOut pools and a token address,
+   * gets the top N pools that involve the other token.
+   * If token address is not on the list, we default to topNSecondHop.
+   * E.g. for a WETH -> USDC swap, if topNTokenInOut found WETH -> DAI and WETH -> USDT,
+   * and there's a mapping USDT => 4, but no mapping for DAI
+   * it would find the top 4 pools that involve USDT, and find the topNSecondHop pools that involve DAI
+   */
+  topNSecondHopForTokenAddress?: Map<string, number>;
```
willpote, line 234: Can we add something to make the key not case sensitive?

willpote (approved):
Left 1 comment, otherwise looks good!

src/routers/alpha-router/alpha-router.ts
mikeki: good point, do we have a function somewhere that normalizes the key?

PR #211, opened 2023-03-21
fix: remove instanceof usage for provider

just-toby:

bug fix

possible versioning mismatch which causes this check to fail when it shouldn't. instead we should use the `isProvider` check on `JsonRpcProvider`

Use the `JsonRpcProvider` built-in function for checking the type.

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -562,11 +562,11 @@ export class AlphaRouter
     }
 
     let gasPriceProviderInstance: IGasPriceProvider;
-    if (this.provider instanceof JsonRpcProvider) {
+    if (JsonRpcProvider.isProvider(this.provider)) {
       gasPriceProviderInstance = new OnChainGasPriceProvider(
         chainId,
-        new EIP1559GasPriceProvider(this.provider),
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -562,11 +562,11 @@ export class AlphaRouter
     }
 
     let gasPriceProviderInstance: IGasPriceProvider;
-    if (this.provider instanceof JsonRpcProvider) {
+    if (JsonRpcProvider.isProvider(this.provider)) {
       gasPriceProviderInstance = new OnChainGasPriceProvider(
         chainId,
-        new EIP1559GasPriceProvider(this.provider),
-        new LegacyGasPriceProvider(this.provider)
+        new EIP1559GasPriceProvider(this.provider as JsonRpcProvider),
```
vm, line 568: is casting necessary here? not sure if this could cause issues / if there's a different way to avoid this

JFrankfurt (approved):

src/routers/alpha-router/alpha-router.ts
just-toby: it was necessary to satisfy the type checker

PR #223, opened 2023-04-13
Update how we fetch percents and amounts for cached routes

mikeki:
Feature/Bug Fix.
Currently we only fetch a quote for the specific percent that the route was cached for, however it's possible that block data is updated and the routes have changed. In such a case it is a good idea to fetch the different percents for all routes to be able to give some margin of error to the caching layer
We will now fetch the different percents for all routes that have been cached
- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1188,9 +1188,12 @@ export class AlphaRouter
     const mixedRoutes = cachedRoutes.routes.filter((route) => route.protocol === Protocol.MIXED);
 
     // Calculate percents from all routes, we will fetch quotes for each percent in case we had stale data when the route was cached
-    const percents: number[] = cachedRoutes.routes.map((route) => route.percent);
-    // If the 100 percent is not included (it was a split route), we will add it back, in case some other route performs better
-    if (!percents.includes(100)) percents.push(100);
+    const percentsSet: Set<number> = new Set(cachedRoutes.routes.map((route) => route.percent));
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1186,20 +1186,25 @@ export class AlphaRouter
     const v3Routes = cachedRoutes.routes.filter((route) => route.protocol === Protocol.V3);
     const v2Routes = cachedRoutes.routes.filter((route) => route.protocol === Protocol.V2);
     const mixedRoutes = cachedRoutes.routes.filter((route) => route.protocol === Protocol.MIXED);
-    const percents: number[] = [];
+
+    // Calculate percents from all routes, we will fetch quotes for each percent in case we had stale data when the route was cached
+    const percentsSet: Set<number> = new Set(cachedRoutes.routes.map((route) => route.percent));
```
ewilz, line 1191: nit: I don't necessarily understand why we're creating a set here only to turn it into an array....I'm not necessarily a typescript wiz so maybe there's a good reason...?

ewilz (approved):
Approving, just wondering if this affects real time latency in prod???

src/routers/alpha-router/alpha-router.ts
mikeki: Mostly to avoid having 100 and 50 twice, so the set will handle duplicates, I opted for that rather than `if (!percents.includes(100)) percents.push(100)`

PR #226, opened 2023-04-20
Add a metric for the quote difference

mikeki:
Metrics update / addition

No metric is logged for the quote difference

We now have a metric for the quote difference

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1001,9 +1001,17 @@ export class AlphaRouter
       const quoteDiff = swapRouteFromChain.quote.subtract(swapRouteFromCache.quote);
       const quoteGasAdjustedDiff = swapRouteFromChain.quoteGasAdjusted.subtract(swapRouteFromCache.quoteGasAdjusted);
       const gasUsedDiff = swapRouteFromChain.estimatedGasUsed.sub(swapRouteFromCache.estimatedGasUsed);
+      const gasAdjustedDiffInAmountTokenTerms = quoteGasAdjustedDiff.multiply(amount.divide(swapRouteFromChain.quoteGasAdjusted));
+      const misquotePercent = gasAdjustedDiffInAmountTokenTerms.divide(amount).multiply(100);
 
       // Only log if diff is not equal to 0
       if (!quoteDiff.equalTo(0)) {
+        metric.putMetric(
+          `TapcompareCachedRoute_quoteGasAdjustedDiffPercent`,
+          Number(misquotePercent.toExact()),
+          MetricLoggerUnit.Percent
+        );
+
         log.warn(
           {
             quoteFromChain: swapRouteFromChain.quote.toExact(),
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1001,9 +1001,17 @@ export class AlphaRouter
       const quoteDiff = swapRouteFromChain.quote.subtract(swapRouteFromCache.quote);
       const quoteGasAdjustedDiff = swapRouteFromChain.quoteGasAdjusted.subtract(swapRouteFromCache.quoteGasAdjusted);
       const gasUsedDiff = swapRouteFromChain.estimatedGasUsed.sub(swapRouteFromCache.estimatedGasUsed);
+      const gasAdjustedDiffInAmountTokenTerms = quoteGasAdjustedDiff.multiply(amount.divide(swapRouteFromChain.quoteGasAdjusted));
+      const misquotePercent = gasAdjustedDiffInAmountTokenTerms.divide(amount).multiply(100);
```
ewilz, line 1005: hard for me to follow whether this is JSBI or BigNumber being used, but I know `BigNumber` doesn't have a concept of decimals, so if these are intended to go below 0, then back up, I think the number would get lost??

ewilz: Ok I see they're CurrencyAmount/Fraction. I feel like there's all kinds of things going on in SOR it can be dizzying (since always haha), I wonder if in general it might be helpful to define types more strictly just for readability.

Just a general thought no need to address now

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1001,9 +1001,17 @@ export class AlphaRouter
       const quoteDiff = swapRouteFromChain.quote.subtract(swapRouteFromCache.quote);
       const quoteGasAdjustedDiff = swapRouteFromChain.quoteGasAdjusted.subtract(swapRouteFromCache.quoteGasAdjusted);
       const gasUsedDiff = swapRouteFromChain.estimatedGasUsed.sub(swapRouteFromCache.estimatedGasUsed);
+      const gasAdjustedDiffInAmountTokenTerms = quoteGasAdjustedDiff.multiply(amount.divide(swapRouteFromChain.quoteGasAdjusted));
```
ewilz, line 1004: if this diff is in "tokenTerms", which terms is quoteGasAdjustedDiff in? Is it not a token?? I'm finding the naming a bit hard to follow

src/routers/alpha-router/alpha-router.ts
ewilz: I personally, as not the biggest math wiz ever, find the calculations and reasoning for them hard to follow, like why must we get misquotePercent with "amountToken" terms.

thinking more strict types and maybe some comments might be helpful

ewilz (approved):
approved, but not easy to follow IMO, maybe some more clarification in code comments or PR description

src/routers/alpha-router/alpha-router.ts
mikeki: thanks for all the feedback, you are right the misquotePercent is a currencyAmount and that's wonky. I'm mostly just trying to do some math here.

What I'm doing is getting the diff between the cache and regular quote, then converting that to the amountIn terms, e.g. if a user is sending 1 ETH, then we get the quote in both methods, we calculate the difference of that quote (which is in tokenOut terms), but to be able to compare against ETH I'm then multiplying the difference by the price of tokenOut with respect to tokenIn (amount/quote).

With that being said, it appears like I can simplify this math.

But essentially what I'm interested in getting is the percent of misquote calculated by the difference and what it would have received in reality

PR #231, opened 2023-04-27
Create BaseAlphaRouter

mikeki:
Refactor

Currently There's AlphaRouter that has a constructor that handles the cases when the params are not provided and uses default params.

Now there's a BaseAlphaRouter that only receives what it really needs as required fields, or optional if they are truly optional.

- **Other information**:


mikeki:
Merging this branch to continue with the iterations, we will still have to review v4.0.0 before we merge into main

PR #234, opened 2023-05-23
Quote the different percents for all routes when coming from cache

mikeki:
feature

currently we use the percents that are in cache

the new behavior will be to quote the percents for all possible routes, just in case we get better quotes for that specific amount

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1196,15 +1196,21 @@ export class AlphaRouter
     const v2Routes = cachedRoutes.routes.filter((route) => route.protocol === Protocol.V2);
     const mixedRoutes = cachedRoutes.routes.filter((route) => route.protocol === Protocol.MIXED);
 
-    // Calculate percents from all routes, we will fetch quotes for each percent in case we had stale data when the route was cached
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1196,15 +1196,21 @@ export class AlphaRouter
...
+    let percents: number[];
+    let amounts: CurrencyAmount[];
+    if (cachedRoutes.routes.length > 1) {
+      // If we have more than 1 route, we will quote the different percents for it, following the regular process
+      [percents, amounts] = this.getAmountDistribution(
+        amount,
+        routingConfig
+      );
+    } else if (cachedRoutes.routes.length == 1) {
+      percents = [100];
```
jsy1218, line 1208: nit: we can use array de-structuring here for consistency:

```
[percents, amounts] = [[100], [amount]];
```

jsy1218:
For my own understanding, even if there are multiple routes, computations are all in-memory right? So that there's no perf impact from this PR.

jsy1218 (approved):

mikeki:
@jsy1218 Computation is in memory, but there might be a perf impact on fetching quotes, since the quotes are fetched (for v3 quotes) by calling the multicall contract

jsy1218 (approved):

PR #237, opened 2023-05-25
feat: [wip] add sepolia

lynnshaoyu:
Adding support for sepolia testnet to Smart Order Router. 

Sepolia command from README finds route successfully from dai to usdc after I added a small dai/usdc liq pool on interface:

Screen Shot 2023-05-31 at 1 29 30 PM


src/util/addresses.ts
```diff
@@ -145,6 +152,13 @@ export const WETH9: {
     'WETH',
     'Wrapped Ether'
   ),
+  [ChainId.SEPOLIA]: new Token(
+    ChainId.SEPOLIA,
+    '0xD0dF82dE051244f04BfF3A8bB1f62E1cD39eED92',
```
cbachmeier, line 157: this is the WETH address that we (Uniswap) deployed: `0xfFf9976782d46CC05630D1f6eBAb18b2324d6B14`

src/util/chains.ts
```diff
@@ -358,6 +376,13 @@ export const WRAPPED_NATIVE_CURRENCY: { [chainId in ChainId]: Token } = {
     'WETH',
     'Wrapped Ether'
   ),
+  [ChainId.SEPOLIA]: new Token(
+    11155111,
+    '0xD0dF82dE051244f04BfF3A8bB1f62E1cD39eED92',
```
cbachmeier, line 381: this is the WETH address that we (Uniswap) deployed: 0xfFf9976782d46CC05630D1f6eBAb18b2324d6B14

lynnshaoyu: TYSM charlie!

src/util/chains.ts
```diff
@@ -65,6 +68,8 @@ export const NETWORKS_WITH_SAME_UNISWAP_ADDRESSES = [
   ChainId.ROPSTEN,
   ChainId.RINKEBY,
   ChainId.GÖRLI,
+  // not sure if this is correct?
+  ChainId.SEPOLIA,
```
zhongeric, line 72: Don't think this is true, this would mean that V3PoolFactory / NFTPositionManager / etc. are deployed on sepolia to the same addresses as mainnet, gorli, etc.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -2571,13 +2565,14 @@ describe('quote for other networks', () => {
       c != ChainId.ARBITRUM_GOERLI &&
       c != ChainId.OPTIMISM && /// @dev infura has been having issues with optimism lately
       // Tests are failing https://github.com/Uniswap/smart-order-router/issues/104
-      c != ChainId.CELO_ALFAJORES
+      c != ChainId.CELO_ALFAJORES &&
+      c != ChainId.SEPOLIA
```
zhongeric, line 2569: Why not run tests on sepolia? If there is reasonable liquidity we should run some tests

zhongeric (approved):
Looks good, just a few comments

src/util/chains.ts
lynnshaoyu: ah ty, removed

test/integ/routers/alpha-router/alpha-router.integration.test.ts
lynnshaoyu: there is not much reasonable liquidity rn :P

PR #238, opened 2023-05-25
BugFix: When symbol is bytes32 we are crashing, this fixes it and handles it gracefully

mikeki:
Bug Fix

Currently if the symbol method is a bytes32 instead of string, we would crash

We don't crash anymore

- **Other information**:
Tested manually with `./bin/cli quote --tokenIn 0x0d88ed6e74bbfd96b831231638b66c05571e824f --tokenOut 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 --amount 1 --exactIn --chainId 1` and got a 200 response with a real quote.

Also added an integration test.

### DIFF
--- a/src/providers/token-provider.ts
+++ b/src/providers/token-provider.ts
@@ -673,7 +673,7 @@ export class TokenProvider implements ITokenProvider {
       } catch (error) {
         log.fatal({ addresses }, `TokenProvider.getTokenSymbol[bytes32] failed with error ${error}.`);
 
-        return Promise.reject(new Error('[TokenProvider.getTokenSymbol] Impossible to fetch token symbol.'));
+        throw new Error('[TokenProvider.getTokenSymbol] Impossible to fetch token symbol.');
       }
     }
 
--- a/src/providers/token-provider.ts
+++ b/src/providers/token-provider.ts
@@ -1,5 +1,5 @@
+import { Interface } from '@ethersproject/abi';
 import { Token } from '@uniswap/sdk-core';
-import { utils } from 'ethers';
 import _ from 'lodash';
 
 import { IERC20Metadata__factory } from '../types/v3/factories/IERC20Metadata__factory';
@@ -644,7 +644,7 @@ export class TokenProvider implements ITokenProvider {
     } catch (error) {
       log.error(`TokenProvider.getTokenSymbol failed with error ${error}. Trying with bytes32.`);
 
-      const bytes32Interface = new utils.Interface([
+      const bytes32Interface = new Interface([
         {
           inputs: [],
           name: 'symbol',
--- a/src/providers/token-provider.ts
+++ b/src/providers/token-provider.ts
@@ -642,7 +642,7 @@ export class TokenProvider implements ITokenProvider {
         providerConfig,
       });
     } catch (error) {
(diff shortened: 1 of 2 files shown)


jsy1218 (approved):

test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -589,6 +581,42 @@ describe('alpha router integration', () => {
           );
         });
 
+        it.only('erc20 -> erc20 works when symbol is returning bytes32', async () => {
```
willpote, line 584: .only

src/providers/token-provider.ts
```diff
@@ -627,6 +628,64 @@ export class TokenProvider implements ITokenProvider {
...
+              name: '',
+              type: 'bytes32',
+            },
+          ],
+          stateMutability: 'view',
+          type: 'function',
+        }
+      ]);
+
+      result = this.multicall2Provider.callSameFunctionOnMultipleContracts<
```
willpote, line 663: Hmm.. thinking through this more, does this mean we retry the whole batch with bytes32? Then wouldn't the other tokens start failing?

zhongeric: Yea this would cause other standard tokens to fail

mikeki: oh boy, that's a good call, there's really no way of knowing if there's one bad token in there that spoils the whole bunch, right?

mikeki: I mean, either way it's gonna fail right? but this would still work for when we are fetching only 1 token and that token is wrong

src/providers/token-provider.ts
```diff
@@ -627,6 +628,64 @@ export class TokenProvider implements ITokenProvider {
...
+      >({
+        addresses,
+        contractInterface: IERC20Metadata__factory.createInterface(),
+        functionName: 'symbol',
+        providerConfig,
+      });
+    } catch (error) {
+      log.error(`TokenProvider.getTokenSymbol failed with error ${error}. Trying with bytes32.`);
+
+      const bytes32Interface = new Interface([
```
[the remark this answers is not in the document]
mikeki: Could you elaborate a bit more here? like what kind of issues do you think could happen if we use the bytes32?

src/providers/token-provider.ts
mikeki: I ran some queries, and the majority of times this runs for only 1 address, so it should work for those cases

src/providers/token-provider.ts
```diff
@@ -660,15 +660,21 @@ export class TokenProvider implements ITokenProvider {
...
+        >({
+          addresses,
+          contractInterface: bytes32Interface,
+          functionName: 'symbol',
+          providerConfig,
+        });
+      } catch (error) {
+        log.fatal({ addresses }, `TokenProvider.getTokenSymbol[bytes32] failed with error ${error}.`);
+
+        return Promise.reject(new Error('[TokenProvider.getTokenSymbol] Impossible to fetch token symbol.'));
```
willpote, line 676: You can just throw new Error - its the same as rejecting the promise if you throw in an async function

willpote (approved):

PR #241, opened 2023-05-30
If symbol comes from bytes32 function, we need to parse it

mikeki:
BugFix
Bytes32 strings are displayed with the raw bytes data
We parse the string
- **Other information**:
Screenshot 2023-05-30 at 12 50 58 PM

### DIFF
--- a/src/providers/token-provider.ts
+++ b/src/providers/token-provider.ts
@@ -1,4 +1,5 @@
 import { Interface } from '@ethersproject/abi';
+import { parseBytes32String } from '@ethersproject/strings';
 import { Token } from '@uniswap/sdk-core';
 import _ from 'lodash';
 
@@ -630,6 +631,7 @@ export class TokenProvider implements ITokenProvider {
 
   private async getTokenSymbol(addresses: string[], providerConfig?: ProviderConfig) {
     let result;
+    let isBytes32 = false;
 
     try {
       result = await this.multicall2Provider.callSameFunctionOnMultipleContracts<
@@ -661,7 +663,7 @@ export class TokenProvider implements ITokenProvider {
       ]);
 
       try {
-        result = this.multicall2Provider.callSameFunctionOnMultipleContracts<
+        result = await this.multicall2Provider.callSameFunctionOnMultipleContracts<
           undefined,
           [string]
         >({
@@ -670,6 +672,7 @@ export class TokenProvider implements ITokenProvider {
           functionName: 'symbol',
           providerConfig,
         });
+        isBytes32 = true;
       } catch (error) {
         log.fatal({ addresses }, `TokenProvider.getTokenSymbol[bytes32] failed with error ${error}.`);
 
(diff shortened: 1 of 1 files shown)


src/providers/token-provider.ts
```diff
@@ -1,4 +1,5 @@
 import { Interface } from '@ethersproject/abi';
+import { parseBytes32String } from '@ethersproject/strings';
```
zhongeric, line 2: What does this func do? strip the tailing 0s off of the bytes string or try converting it into ASCII?

src/providers/token-provider.ts
```diff
@@ -630,6 +631,7 @@ export class TokenProvider implements ITokenProvider {
 
   private async getTokenSymbol(addresses: string[], providerConfig?: ProviderConfig) {
```
zhongeric, line 632: Can we add a type to the return value(s) of this func?

src/providers/token-provider.ts
```diff
@@ -730,7 +734,11 @@ export class TokenProvider implements ITokenProvider {
           continue;
         }
 
-        const symbol = symbolResult.result[0]!;
+        let symbol = symbolResult.result[0]!;
```
jsy1218, line 737: nit: probably better to retain 'const' variable, rather than re-assign (e.g. via ternary operator):

```
const symbol = isBytes32 ? parseBytes32String(symbolResult.result[0]!) : symbolResult.result[0]!;
```

jsy1218 (approved):

src/providers/token-provider.ts
mikeki: this is the JS function:
```
function parseBytes32String(bytes) {
    var data = (0, bytes_1.arrayify)(bytes);
    // Must be 32 bytes with a null-termination
    if (data.length !== 32) {
        throw new Error("invalid bytes32 - not 32 bytes long");
    }
    if (data[31] !== 0) {
        throw new Error("invalid bytes32 string - no null terminator");
    }
    // Find the null termination
    var length = 31;
    while (data[length - 1] === 0) {
        length--;
    }
    // Determine the string value
    return (0, utf8_1.toUtf8String)(data.slice(0, length));
}
```

it appears like the `toUtf8String` does the conversion

src/providers/token-provider.ts
mikeki: thanks good suggestion!

src/providers/token-provider.ts
jsy1218: bcc from the library:

```
export function parseBytes32String(bytes: BytesLike): string {
    const data = arrayify(bytes);

    // Must be 32 bytes with a null-termination
    if (data.length !== 32) { throw new Error("invalid bytes32 - not 32 bytes long"); }
    if (data[31] !== 0) { throw new Error("invalid bytes32 string - no null terminator"); }

    // Find the null termination
    let length = 31;
    while (data[length - 1] === 0) { length--; }

    // Determine the string value
    return toUtf8String(data.slice(0, length));
}
```

so 1) truncating trailing 0s 2) convert to utf8.

PR #242, opened 2023-05-31
3.11.6

mikeki:
Release containing bugfix addressed in #241 

Token Symbols obtained from Bytes32 are not parsed and currently display the Hex value.

In 3.11.6 the Token symbols that are obtained from a Bytes32 function will be parsed into their UTF values

- **Other information**:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@uniswap/smart-order-router",
-  "version": "3.11.5",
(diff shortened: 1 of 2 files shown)


cgkol (commented):
I think it would be nice in the PR description to explain which changes are included with this release (i think most of us dont remember where the last one was cut etc)

cgkol (approved):

PR #248, opened 2023-06-09
[ROUTE-9] pass down block number to v3 static subgraph provider

jsy1218:
Bug fix

`StaticV3SubgraphProvider` currently doesn't get the block number from the alpha router to get the pools.

`StaticV3SubgraphProvider` gets the block number from the alpha router, and passes down to the downstream pool provider (concrete instance of `IV3PoolProvider`) to get the pools.

- **Other information**:
This is to complement the routing-api pr [211](https://github.com/Uniswap/routing-api/pull/211). Dynamo caching pool provider will always be cache miss until we pass down the block number.


mikeki:
LGTM

mikeki (approved):

cgkol (approved):

PR #258, opened 2023-06-13
chore: remove deprecated testnets

lynnshaoyu:
also uses ChainId from sdk-core instead.


src/util/chains.ts
```diff
@@ -1,78 +1,48 @@
...
+// TO REMOVE: 
+// export enum ChainId {
+//   ROPSTEN = 3,
+//   RINKEBY = 4,
+//   KOVAN = 42,
+//   OPTIMISTIC_KOVAN = 69,
+//   ARBITRUM_RINKEBY = 421611,
 
 // WIP: Gnosis, Moonbeam
 export const SUPPORTED_CHAINS: ChainId[] = [
```
mikeki, line 12: Let's also update this in routing-api

jsy1218:
Looks like this PR forgot to check in the `package-lock.json`. Supposedly the local has run `npm install`

PR #261, opened 2023-06-14
feat: use sdk core ChainId and add avax

lynnshaoyu:


package-lock.json
```diff
@@ -12,6 +12,7 @@
         "@uniswap/default-token-list": "^2.0.0",
         "@uniswap/permit2-sdk": "^1.2.0",
         "@uniswap/router-sdk": "^1.3.0",
+        "@uniswap/sdk-core": "^3.2.6",
```
jsy1218, line 15: The amount of changes in `package-lock.json` is way too high. I expect much smaller number of changes. With this many dependencies changes, we might have unexpected side effects when getting the newer sor adopted in [interface](https://github.com/Uniswap/interface) and [routing-api](https://github.com/Uniswap/routing-api)

package-lock.json
```diff
@@ -3170,9 +3347,9 @@
       }
     },
     "node_modules/@uniswap/sdk-core": {
```
jsy1218, line 3349: sor already has transitive dependency on sdk-core through multiple packages, v2-sdk, v3-sdk, etc. All follows the semantic versioning with the caret (^) prefix. I'm wondering if you only change the sdk-core version in `package-lock.json`, would it work?

package.json
```diff
@@ -34,6 +34,7 @@
     "@uniswap/default-token-list": "^2.0.0",
     "@uniswap/permit2-sdk": "^1.2.0",
     "@uniswap/router-sdk": "^1.3.0",
+    "@uniswap/sdk-core": "^3.2.6",
```
jsy1218, line 37: Revisit on adding the sdk-core explicitly since we already have transitive dependency.

jsy1218 (changes requested):

package-lock.json
lynnshaoyu: good idea, I fixed it by adding an overrides part to package.json, to force the transitive dependency of sdk-core to be version 3.2.6.

jsy1218 (approved):

PR #262, opened 2023-06-15
[fix] [ROUTE-9] optimize TVL heuristics calculation paths (aka RPC calls) across v3 and mixed quotes

jsy1218:
Enhancement

After we eliminate the dynamo pool caching miss due to missing block number (https://github.com/Uniswap/smart-order-router/pull/253 & https://github.com/Uniswap/smart-order-router/pull/257), we still see non-trivial amount of cache miss due to not in the table (meaning the block number all passed down to the dynamo pool caching provider now):
Screenshot 2023-06-14 at 8 22 34 AM
Screenshot 2023-06-14 at 8 22 39 AM

I was trying to repro this in my personal CDK stack and realized because v3 and mixed quoters call the same [getHighestLiquidityV3USDPool](https://github.com/Uniswap/smart-order-router/blob/main/src/util/gas-factory-helpers.ts#L100)  and [getHighestLiquidityV3NativePool](https://github.com/Uniswap/smart-order-router/blob/main/src/util/gas-factory-helpers.ts#L54). Because of the async timing, even with in-memory cache, along with the main [DynamoDBCachingV3PoolProvider](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/pools/pool-caching/v3/dynamo-caching-pool-provider.ts) algorithm, it might still be cache miss. [DynamoDBCachingV3PoolProvider](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/pools/pool-caching/v3/dynamo-caching-pool-provider.ts) algorithm is such that it tries to get pool individually from the dynamo caching, and then for each cache miss, we tally and try to backfill after a single getPools call from the multicall contract. The time begin get cache and set set is a bit apart because of this algorithm, specifically the multicall contract. However, we don't want to change this algorithm, because we want to only call the multicall contract once for all the given pools. In my personal AWS log insights, I can see thus the bad timing that two dynamo cache get are ahead of the two dynamo cache set:

Screenshot 2023-06-15 at 8 27 34 AM

The new behavior is we perform the following TVL heuristics only once across v3 and mixed quotes:

1. USD pools max liquidity
2. Native against quote token pools max liquidity
3. Native against amount token pools max liquidity

Then we pass those calculated max liquidity pools into the m3 and mixed quotes gas model path.

- **Other information**:
After personal cdk testing against routing-api, I feel confident this is the right optimization (and they are apple-to-apple comparison, no logging, no dynamo config tuning, and locally adjusted to sample at 100%):

1. Screenshot 1 [ran](https://uniswapteam.slack.com/archives/C05A17SKC0M/p1686765688984989) from yesterday 10:30-10:40am PST - We see the peak cache hit at 60%
Screenshot 2023-06-15 at 10 58 59 AM

2. Screenshot 2 ran this morning 10:20-10:30am PST - We see the peak cache hit at 75%
Note the test call pattern is that we test against different chains and exact in and out sequentially, so the max local cache hit rate will be a bit lower than in prod. In prod, I expect to see more effective cache hit due to the repeated quotes with the same parameters (quote vs amount tokens, chains etc).
Screenshot 2023-06-15 at 10 59 06 AM

Note the test call pattern is that we test against different chains and exact in and out sequentially, so the max local cache hit rate will be a bit lower than in prod. In prod, I expect to see more effective cache hit due to the repeated quotes with the same parameters (quote vs amount tokens, chains etc).

### DIFF
--- a/src/routers/alpha-router/gas-models/gas-model.ts
+++ b/src/routers/alpha-router/gas-models/gas-model.ts
@@ -34,7 +34,6 @@ import {
 } from '../../../providers/token-provider';
 import { IV2PoolProvider } from '../../../providers/v2/pool-provider';
 import { ArbitrumGasData, IL2GasDataProvider, OptimismGasData, } from '../../../providers/v3/gas-data-provider';
-import { IV3PoolProvider } from '../../../providers/v3/pool-provider';
 import { CurrencyAmount } from '../../../util/amounts';
 import { ChainId } from '../../../util/chains';
 import {
@@ -43,6 +42,7 @@ import {
   V2RouteWithValidQuote,
   V3RouteWithValidQuote,
 } from '../entities/route-with-valid-quote';
+import { Pool } from '@uniswap/v3-sdk';
 
 export const usdGasTokensByChain: { [chainId in ChainId]?: Token[] } = {
   [ChainId.MAINNET]: [DAI_MAINNET, USDC_MAINNET, USDT_MAINNET],
@@ -74,7 +74,7 @@ export type L1ToL2GasCosts = {
 export type BuildOnChainGasModelFactoryType = {
   chainId: ChainId;
   gasPriceWei: BigNumber;
-  v3poolProvider: IV3PoolProvider;
+  pools: LiquidityCalculationPools;
   amountToken: Token;
   quoteToken: Token;
   v2poolProvider: IV2PoolProvider;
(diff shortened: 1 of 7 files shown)


src/routers/alpha-router/gas-models/gas-model.ts
```diff
@@ -91,6 +91,12 @@ export type BuildV2GasModelFactoryType = {
   token: Token;
 };
 
+export type LiquidityCalculationPools = {
+  usdPool: Pool
+  nativeQuoteTokenV3Pool: Pool | null
+  nativeAmountTokenV3Pool: Pool | null
```
mikeki, line 97: these can be simplified as:
```
  nativeQuoteTokenV3Pool?: Pool
  nativeAmountTokenV3Pool?: Pool
```

zhongeric (approved):

PR #278, opened 2023-07-14
chore: update dependencies

just-toby:
pull the newest versions of shared SDKs


package-lock.json
jsy1218, line 1: The number of changes in `package-lock.json` is too high. I used the same `package.json` change and the attached is the `package-lock.json` diff from my local patch.

[shared-sdk.patch](https://github.com/Uniswap/smart-order-router/files/12053807/shared-sdk.patch)

My local node version is v18.13.0 and npm version is 8.19.3

src/providers/v3/static-subgraph-provider.ts
```diff
@@ -76,64 +75,61 @@ const BASES_TO_CHECK_TRADES_AGAINST: ChainTokenList = {
...
     WRAPPED_NATIVE_CURRENCY[ChainId.GOERLI]!,
     USDT_GOERLI,
     USDC_GOERLI,
     WBTC_GOERLI,
-    DAI_GOERLI,
-  ],
-  [ChainId.SEPOLIA]: [
-    WRAPPED_NATIVE_CURRENCY[ChainId.SEPOLIA]!,
-    USDC_SEPOLIA,
+    DAI_GOERLI
```
jsy1218, line 85: nice catch...

package-lock.json
just-toby: thanks

jsy1218 (approved):

PR #281, opened 2023-07-18
feat: fix decimal issue blocking avax

lynnshaoyu:
tested locally for avax:
Screen Shot 2023-07-17 at 9 03 46 PM


jsy1218 (changes requested):
Discussed in the slack channel, we want to minimize invasive changes during ethCC, and thus prefer my one-line fix https://github.com/Uniswap/smart-order-router/pull/282.

PR #286, opened 2023-07-20
feat: make USD gas estimation optional

just-toby:

Default to a USD Gas Estimation of 0  if there are no USD-type Token pools.

Currently, this package throws an Error if there are no USD-type Token pools specified.

If we don't have a USD pool, we just fallback to a gas estimation of zero.

- **Other information**:

### Test Plan
I was able to run the unit tests successfully, but couldn't get the integ-tests to run on my machine even with the .env file populated.

I'm also not sure where exactly to add a unit test for this new behavior, so looking for guidance on that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@uniswap/smart-order-router",
-  "version": "3.13.6",
+  "version": "3.13.7",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@uniswap/smart-order-router",
-      "version": "3.13.5",
+      "version": "3.13.7",
       "license": "GPL",
       "dependencies": {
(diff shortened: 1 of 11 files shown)


just-toby:
> Curious - what's the motivation for this change-set @just-toby?
> 
> Looking at these gas models, we've got no unit test coverage here as it stands. I'd love to get some initial test coverage down before we make changes.

@andysmith415 right now the entire functionality of routing is blocked by being able to calculate the USD gas estimation, this PR is trying to see if that's necessary. for context, i'm trying to see if we can build client-side routing support for BASE network, which doesn't have any stablecoins yet.

without this, we would just need to wait for the stablecoins to be deployed, at which point we'd have API-side routing anyway

just-toby:
discussed offline with @mikeki and we agreed that it would be better to stick with the status quo and keep the stablecoin requirement

PR #289, opened 2023-07-25
Adjust gas estimates for single hop, and reduce cache length

willpote:

Two modifications to the algorithm we use for computing gas estimates: 
1/ increase estimate for single hop swaps.
2/ add a custom overhead for certain known expensive to transfer tokens (AAVE and LDO)
Also reduces cache length for the gas price from 15s -> 7s

Gas prices are cached for 15s
There is no functionality for applying gas overheads based on the tokens being swapped
There is no gas overhead applied for single hop swaps

Cache gas for 7s
Apply a 15k overhead to the route if it has 1 hop
Add a 150k overhead for routes that involve AAVE or LDO

- **Other information**:

Making this change based on a few pieces of data / thoughts
1/ https://uniswapteam.slack.com/archives/C021SU4PMR7/p1688666362630569 
2/ Multiple UniswapX market makers have stated that they feel we are underestimated gas costs to fill orders
3/ Generally it is best to overestimate than underestimate, both for the user (no surprises) and for UniswapX (if we are underestimating gas on classic swaps too much we end up sending much fewer swaps to X)
4/ Caching gas prices for longer than 1 block (12s) could lead to stale gas data being used. Better to shorten it. Also the 15s value was picked before PoS. Now that blocks are fairly consistently 12s, 15s is too long
5/ AAVE and LDO are well known to have expensive transferFrom functions. This has been well known for a long time (see https://www.notion.so/uniswaplabs/Improving-Gas-Estimates-in-the-Auto-Router-bd1982e86271429dbb24f436aebeb5ba?pvs=4#e7d3a0827027437ba550a0b726085fd4) but we have never taken action. This is now causing issues in UniswapX, as we parameterize X orders using the gasAdjustedOutput. The underestimate is causing us to price these tokens too highly

### DIFF
--- a/src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
+++ b/src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
@@ -13,9 +13,7 @@ import {
   OptimismGasData,
 } from '../../../../providers/v3/gas-data-provider';
 import { CurrencyAmount } from '../../../../util/amounts';
-import {
-  getL2ToL1GasUsed,
-} from '../../../../util/gas-factory-helpers';
+import { getL2ToL1GasUsed } from '../../../../util/gas-factory-helpers';
 import { log } from '../../../../util/log';
 import {
   buildSwapMethodParameters,
@@ -33,6 +31,7 @@ import {
   COST_PER_HOP,
   COST_PER_INIT_TICK,
   COST_PER_UNINIT_TICK,
+  SINGLE_HOP_OVERHEAD,
 } from './gas-costs';
 
 /**
@@ -64,7 +63,7 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
     pools,
     amountToken,
     quoteToken,
-    l2GasDataProvider
+    l2GasDataProvider,
   }: BuildOnChainGasModelFactoryType): Promise<
     IGasModel<V3RouteWithValidQuote>
   > {
@@ -89,10 +88,7 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
       };
       let l1Used = BigNumber.from(0);
       let l1FeeInWei = BigNumber.from(0);
-      if (
-        chainId == ChainId.OPTIMISM ||
-        chainId == ChainId.OPTIMISM_GOERLI
-      ) {
+      if (chainId == ChainId.OPTIMISM || chainId == ChainId.OPTIMISM_GOERLI) {
         [l1Used, l1FeeInWei] = this.calculateOptimismToL1SecurityFee(
           route,
           swapOptions,
@@ -367,7 +363,14 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
     );
     const totalHops = BigNumber.from(routeWithValidQuote.route.pools.length);
 
-    const hopsGasUse = COST_PER_HOP(chainId).mul(totalHops);
+    let hopsGasUse = COST_PER_HOP(chainId).mul(totalHops);
+
+    // We have observed that this algorithm tends to underestimate single hop swaps.
+    // We add a buffer in the case of a single hop swap.
+    if (totalHops.eq(1)) {
+      hopsGasUse = hopsGasUse.add(SINGLE_HOP_OVERHEAD(chainId));
+    }
+
     const tickGasUse = COST_PER_INIT_TICK(chainId).mul(
       totalInitializedTicksCrossed
     );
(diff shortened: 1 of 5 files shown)


cgkol (approved):
NIt: The pr descriptioon says the cache is still 15 secs, we should move it to 7;

Change looks good code-wise, i am still unsure whats the rationale behind increasing only for single hops and not all of them; any corroborating data in that regard? 

the only message i see is this one https://uniswapteam.slack.com/archives/C021SU4PMR7/p1688666876474179?thread_ts=1688666362.630569&cid=C021SU4PMR7

jsy1218:
How urgent is it to merge this PR? Right now the [alpha-router.integration.test.ts](https://github.com/Uniswap/smart-order-router/blob/main/test/integ/routers/alpha-router/alpha-router.integration.test.ts) runs with test failures. I'm hoping we can fix the regressions/test failures before we merge any more PR.

mikeki (changes requested):

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -367,7 +363,14 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
...
 
-    const hopsGasUse = COST_PER_HOP(chainId).mul(totalHops);
+    let hopsGasUse = COST_PER_HOP(chainId).mul(totalHops);
+
+    // We have observed that this algorithm tends to underestimate single hop swaps.
+    // We add a buffer in the case of a single hop swap.
+    if (totalHops.eq(1)) {
+      hopsGasUse = hopsGasUse.add(SINGLE_HOP_OVERHEAD(chainId));
+    }
+
```
mikeki, line 373: Could we instead add the overhead to every calculation regardless of wether it is a single hop or not? the special case feel strange too.

willpote: We've only observed the gas being underestimated for single hops, multiple hops seem much closer, so don't think we should apply it to everything

willpote:
> NIt: The pr descriptioon says the cache is still 15 secs, we should move it to 7;
> 
> Change looks good code-wise, i am still unsure whats the rationale behind increasing only for single hops and not all of them; any corroborating data in that regard?
> 
> the only message i see is this one https://uniswapteam.slack.com/archives/C021SU4PMR7/p1688666876474179?thread_ts=1688666362.630569&cid=C021SU4PMR7

We don't have extensive data for this, mostly just what we've been hearing from market makers, and what we've been seeing pretty consistently anecdotally during Gouda testing

The whole gas estimation algorithm is heuristic based anyway and purely based on instinct and observations. Would love to have something more scientific but we aren't there yet, and won't be for a while

willpote:
> If we're touching this code, I'd love to add some unit test coverage. Without any coverage beyond the integration tests, it's hard to understand the impact of this change.
> 
> I left a similar comment on a recent PR (https://github.com/Uniswap/smart-order-router/pull/286/files#diff-2dad352ff1a3fdbb82621ff1c08a0799a77d76282a106b8a753042d785691689).

Setting up unit tests here would take quite a bit of time, and I have a bit of time pressure to get these in as we are actively seeing issues with orders on X because of this. 

FWIW I think this is one of the lower risk un-unit-tested places in the code, as a gas estimate is just a rough estimate, and the whole thing is heuristic based anyways

willpote:
> How urgent is it to merge this PR? Right now the [alpha-router.integration.test.ts](https://github.com/Uniswap/smart-order-router/blob/main/test/integ/routers/alpha-router/alpha-router.integration.test.ts) runs with test failures. I'm hoping we can fix the regressions/test failures before we merge any more PR.

This is causing issues in X, so quite urgent now!

jsy1218:
> > How urgent is it to merge this PR? Right now the [alpha-router.integration.test.ts](https://github.com/Uniswap/smart-order-router/blob/main/test/integ/routers/alpha-router/alpha-router.integration.test.ts) runs with test failures. I'm hoping we can fix the regressions/test failures before we merge any more PR.
> 
> This is causing issues in X, so quite urgent now!

Got it. We actually have [leads](https://uniswapteam.slack.com/archives/C021SU4PMR7/p1690585792407429?thread_ts=1690565283.482299&cid=C021SU4PMR7) on why integ-tests are failing, so we can probably unblock this urgent fix, and fast follow-up in fixing the integ-tests.

mikeki (approved):
Let's try to add those unit tests as soon as possible, if it's within this branch great, if not then let's do it before we forget.

PR #290, opened 2023-07-25
fix integ-test

jsy1218:
Bug fix.

Integ-tests are failing at runtime, because we don't have stablecoins on BASE and BASE_GOERLI yet.
In addition, we skipped avalanche for testing

We hard-code the BASE and BASE_GOERLI to be avalanche one, but skip BASE and BASE_GOERLI for now.

- **Other information**:
Somehow my local can't run BNB and AVAX chain tests successful. Want to see if others can run successfully on their local.

### DIFF
--- a/test/integ/routers/alpha-router/alpha-router.integration.test.ts
+++ b/test/integ/routers/alpha-router/alpha-router.integration.test.ts
@@ -2559,8 +2559,9 @@ describe('quote for other networks', () => {
     [ChainId.MOONBEAM]: WBTC_MOONBEAM,
     [ChainId.BNB]: USDC_BNB,
     [ChainId.AVALANCHE]: USDC_ON(ChainId.AVALANCHE),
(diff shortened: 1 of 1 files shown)


cgkol (commented):
If we are going to skip base anyway, do we need to change line 2583 and line 2563>

jsy1218:
> I agree with @cgkol here, is it possible to avoid those changes?

@cgkol @mikeki Unfortunately, there will be runtime error, if we don't make those changes. The const variables make it evaluate all the ChainId during the runtime, so for BASE and BASE_GOERLI the test is bound to throw exception under 

```
describe('quote for other networks', () => {...})
```

https://github.com/Uniswap/smart-order-router/blob/main/src/providers/token-provider.ts#L750

jsy1218:
Offline discussed, we decided to abandon this PR, and have another attempt to fix for all for good.

PR #291, opened 2023-07-27
Make TEST_ERC20_1 and _2 lazily evaluated

lynnshaoyu:

- **Other information**:


test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -536,7 +536,7 @@ describe('alpha router integration', () => {
    */
   for (const tradeType of [TradeType.EXACT_INPUT, TradeType.EXACT_OUTPUT]) {
     describe(`${ID_TO_NETWORK_NAME(1)} alpha - ${tradeType}`, () => {
-      describe(`+ Execute on Hardhat Fork`, () => {
+      describe.skip(`+ Execute on Hardhat Fork`, () => {
```
jsy1218, line 539: Skipping hardhat fork is not the right step in a direction. We know which commit caused the hardhat fork tests to start failing, and we should fix there.

PR #300, opened 2023-08-10
fix: buggy max by sorting for liquidity among pools

zhongeric:
TLDR: there is a bug in how the pool with the highest liquidity is chosen amount v3 pools. This was causing us to use a very illiquid 1inch<>WETH pool for quoting gas prices, leading to a severely low gas estimation. 

This fixes it, though we should add a task to refactor this logic / design

### DIFF
--- a/test/unit/routers/alpha-router/util/gas-factory-helpers.test.ts
+++ b/test/unit/routers/alpha-router/util/gas-factory-helpers.test.ts
@@ -22,7 +22,7 @@ const mockUSDCNativePools = [
   USDC_WETH_HIGH_LIQ_HIGH,
 ];
 
-describe.only('getHighestLiquidity pool tests', () => {
+describe('getHighestLiquidity pool tests', () => {
   let mockPoolProvider: sinon.SinonStubbedInstance<V3PoolProvider>;
 
   beforeEach(() => {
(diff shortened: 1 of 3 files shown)


mikeki (dismissed):

jsy1218:
routing-api team discussed to start adding more unit test coverage against [smart-order-router](https://github.com/Uniswap/smart-order-router). This seems like a good use of it. I think the team can add it as a fast follow-up, so that we don't block merging this PR. It's probably affecting many v3 route.

jsy1218 (dismissed):

rileydcampbell (dismissed):

mikeki (approved):

jsy1218 (dismissed):

test/unit/routers/alpha-router/util/gas-factory-helpers.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  USDC_WETH_MED_LIQ_MEDIUM,
+} from '../../../../test-util/mock-data';
+
+const mockUSDCNativePools = [
+  USDC_WETH_LOW_LIQ_LOW,
+  USDC_WETH_MED_LIQ_MEDIUM,
+  USDC_WETH_HIGH_LIQ_HIGH,
+];
+
+describe.only('getHighestLiquidity pool tests', () => {
```
mikeki, line 25: `.only` here

mikeki (changes requested):
Sorry, taking back the previous approval.

I was wondering if you have a test case that is targetted towards the specific bug that you just solved.

One that would return the incorrect pool when using the `maxBy` method and the right pool when using your bug fix.

mikeki (approved):

PR #303, opened 2023-08-12
fix: hardhat fork in the integ-tests

jsy1218:
Bug fix

All the hardfork integ-tests are failing. See the warp link under https://github.com/Uniswap/smart-order-router/pull/299.

We use a later block, that has the universal router v1.2 deploy address. And for WETH whale, we need to change the address where the new address has at least 4000 WETH with the newly hard-coded block.

- **Other information**:
For some reason, base mainnet integ-tests start failing with insufficient gas limits https://app.warp.dev/block/tCQ8sTuuw42vtQjttnDlAE. But we can follow-up in a linear ticket.

See more context on why this fix in the slack [thread](https://uniswapteam.slack.com/archives/C021SU4PMR7/p1690932052416269?thread_ts=1690565283.482299&cid=C021SU4PMR7). (A bit convoluted, not going to illustrate here..)

### DIFF
--- a/test/integ/routers/alpha-router/alpha-router.integration.test.ts
+++ b/test/integ/routers/alpha-router/alpha-router.integration.test.ts
@@ -73,7 +73,7 @@ import { Permit2__factory } from '../../../../src/types/other/factories/Permit2_
 import { getBalanceAndApprove } from '../../../test-util/getBalanceAndApprove';
 import { WHALES } from '../../../test-util/whales';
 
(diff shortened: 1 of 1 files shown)


mikeki (approved):
Can we enable integration tests in GitHub now? I think we just have to add the secrets

jsy1218:
> Can we enable integration tests in GitHub now? I think we just have to add the secrets

Exactly, yes we can

PR #304, opened 2023-08-12
fix base multicall multichunk for gas

jsy1218:
Bug fix

Base mainnet integ-tests are failing due to gas https://app.warp.dev/block/tCQ8sTuuw42vtQjttnDlAE.

Reduce Base multichunk so swap integ-tests aren't failing anymore. https://app.warp.dev/block/qBFFXAPgnzIsqjX4atI7Pd

- **Other information**:
Not sure when we started to see BASE gas failures. But it might impact prod, as we started seeing the gas error [metrics](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#metricsV2?graph=~(metrics~(~(~'Uniswap~'QuoteOutOfGasExceptionRetry~'Service~'RoutingAPI))~view~'timeSeries~stacked~false~region~'us-east-2~start~'-PT8640H~end~'P0D~stat~'Sum~period~300)&query=~'*7bUniswap*2cService*7d*20GAS) recently:
Screenshot 2023-08-11 at 9 43 54 PM

The timeline doesn't really match, but we could keep an eye out.

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -388,7 +388,6 @@ export class AlphaRouter
       switch (chainId) {
         case ChainId.OPTIMISM:
         case ChainId.OPTIMISM_GOERLI:
-
           this.onChainQuoteProvider = new OnChainQuoteProvider(
             chainId,
             provider,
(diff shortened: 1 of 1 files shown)


mikeki (approved):
We should also update the routing-api config if this config works better

jsy1218:
> We should also update the routing-api config if this config works better

I just realized that we don't override `BASE` gas params in routing-api 😅 https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L177-L246

PR #305, opened 2023-08-13
add integ test into github workflow

jsy1218:
Enhancement

Github workflow test only runs unit tests, but not integration tests

Add integration tests into the github workflow

- **Other information**:
[TODO] - add remaining JSON RPC URLs into this github repo secrets. Require someone with [smart-order-router](https://github.com/Uniswap/smart-order-router) repo admin access.
In case we have to ship hotfix asap, we can skip the workflow run https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs


.github/workflows/tests.yml
```diff
@@ -8,14 +8,14 @@ on:
...
-    name: Unit Tests
+    name: Unit Tests & Integration Tests
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 12.x
+          node-version: 18.x
```
zhongeric, line 18: why the large node bump to 18? what about 16

zhongeric (approved):
lgtm, one thing for why we didn't include integ tests in the workflow initially was that it would make all PRs take 30 min to pass actions

.github/workflows/tests.yml
jsy1218: Because both routing-api and interface node env are on v18 now.

mikeki (approved):

PR #309, opened 2023-08-17
Add cached routes routing configs to control the usage of the cache

mikeki:
feature

currently there's no configuration to control the usage of the cache, or overwrite it

The new configuration options allow us to control from routing api wether the cache will be used, and writte to, or not.

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -313,6 +313,16 @@ export type AlphaRouterConfig = {
    * 40% of input => Route 3
    */
   distributionPercent: number;
+  /**
+   * Flag to indicate whether to use the cached routes or not.
+   * By default, the cached routes will be used.
+   */
+  useCachedRoutes?: boolean;
+  /**
+   * Flag to indicate whether to write to the cached routes or not.
+   * By default, the cached routes will be written to.
+   */
+  writeToCachedRoutes?: boolean;
 };
 
 export class AlphaRouter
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -909,7 +919,11 @@ export class AlphaRouter
     const blockNumber = partialRoutingConfig.blockNumber ?? this.getBlockNumberPromise();
 
     const routingConfig: AlphaRouterConfig = _.merge(
-      {},
+      {
+        // These settings could be changed by the partialRoutingConfig
+        useCachedRoutes: true,
```
jsy1218, line 924: Looks like we will need to explicitly override to true in routing-api as well, if we were to implement this way. I'm thinking if we can do the opposite:

```
dontUseCachedRoutes?: boolean
dontWriteToCachedRoutes?: boolean;

routingConfig.dontUseCachedRoutes && !routingConfig.dontUseCachedRoutes
...
```

On the downside, there's even more cognitive load to this complex system

jsy1218 (approved):

src/routers/alpha-router/alpha-router.ts
jsy1218: nvm, I interpreted wrong. left-side of the lodash merge is `useCachedRoutes: true` and `writeToCachedRoutes: true` and right-side by default is `useCachedRoutes: undefined` and `writeToCachedRoutes: undefined`, so it takes the defined values.

PR #316, opened 2023-08-18
Chore: measure per protocol version getRoutesThenQuotes and getQuotes latencies

jsy1218:
Add latency instrumentation

We don't know the overall per-protocol version quotes latencies.

We will measure the overall per-protocol version quotes latencies.

- **Other information**:
The latencies breakdown we have might miss something. In order to do better fast quote analysis, we need to see the per-protocol overall latencies.
Tested in quoting-api via https://github.com/Uniswap/routing-api/compare/main...jsy1218/explore-sor-local-tarball:
Screenshot 2023-08-18 at 4 01 30 PM

### DIFF
--- a/src/routers/alpha-router/quoters/base-quoter.ts
+++ b/src/routers/alpha-router/quoters/base-quoter.ts
@@ -5,7 +5,13 @@ import { Pool } from '@uniswap/v3-sdk';
 import _ from 'lodash';
 
 import { ITokenListProvider, ITokenProvider, ITokenValidatorProvider, TokenValidationResult } from '../../../providers';
-import { CurrencyAmount, log, poolToString } from '../../../util';
+import {
+  CurrencyAmount,
+  log,
+  metric,
+  MetricLoggerUnit,
+  poolToString
+} from '../../../util';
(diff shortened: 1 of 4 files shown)


src/routers/alpha-router/quoters/base-quoter.ts
```diff
@@ -107,19 +114,41 @@ export abstract class BaseQuoter<Route extends V2Route | V3Route | MixedRoute> {
...
+            percents,
+            quoteToken,
+            tradeType,
+            routingConfig,
+            routesResult.candidatePools,
+            gasModel,
+            gasPriceWei
+          ).then((quotesResult) => {
+            metric.putMetric(
+              `${this.quoterVersion}GetQuotesLoad`,
```
mikeki, line 135: do we know that this metric will not collide with the other existing metric of similar name?

jsy1218: Good point. Although the existing metric is called `*QuotesLoad`, it might be confusing when we add this one. I think we can name it something like `*OverallGetQuotesLoad`, wdyt?

mikeki (approved):

PR #323, opened 2023-08-20
Measure per-protocol quote latencies and request counts for both cached routes and non-cached routes

jsy1218:
Add per-protocol and cached vs non-cached latencies and request count instrumentation

With https://github.com/Uniswap/smart-order-router/pull/316, we did gain a bit of insights on the per-protocol `getRoutesThenQuotes` latencies breakdown. However this alone is not good enough for us to see where we can most effectively shave off the fast quote latencies. This is because the metrics being added mostly correspond to the non-cached routes getting routes and quotes path. With cached routes, we want to understand the request count distribution of cached routes vs. non-cached routes, and then within each one, we then want to understand the per-protocol breakdown. We want to see whether it's more effective to focus on shave off cached routes or non-cached routes first. Then because getting quotes is known to be a bottleneck (because for v2, we constantly reload reserves without caching at all in routing-api only, and for v3, the currently very low distribution percent (5%) caused the on-chain quote provider to get the multiple of 20 quotes per discovered route), we want to see is it v2 or v3 or mixed that's mostly bottleneck the quote latencies. This is important because interface sends `v2,v3,mixed` for all quotes, and we use `Promise.all` to parallelize all the calls.

Screenshot 2023-08-18 at 4 26 31 PM

We added the per-protocol break down and cached vs non-cached latencies and request volume instrumentation.

We can breakdown sor algorithm to following couple high-levels (true for v2, v3, as well as mixed):

1. routes discovery algorithm
2. routes getting valid quotes algorithm
3. calculate best swap algorithm from routes with valid quotes.

For (1), it's optimized by cached routes. But we don't know how many quotes requests from interface hits the cached routes. For (2), we already know v2 getting reserves don't have cache, and v3 has very low distribution percent. For (3), meanwhile the quotes latencies can be impacted by graph complexities, view (3) as a complete in-memory Dijkstra's Shortest Path Algorithm (within `getBestSwapRoute`), and hence less effective in shaving off the quote latencies.

- **Other information**:
Tested via local tarball method, and can see metrics in my local routing-api:
Screenshot 2023-08-20 at 2 35 02 PM

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1279,6 +1279,9 @@ export class AlphaRouter
     if (v3Routes.length > 0) {
       const v3RoutesFromCache: V3Route[] = v3Routes.map((cachedRoute) => cachedRoute.route as V3Route);
 
+      const beforeGetQuotes = Date.now();
+      metric.putMetric("SwapRouteFromCache_V3_GetQuotes_Request", 1, MetricLoggerUnit.Count);
+
       quotePromises.push(
         this.v3Quoter.getQuotes(
           v3RoutesFromCache,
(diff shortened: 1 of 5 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1307,13 +1320,24 @@ export class AlphaRouter
...
+          return result;
+        })
       );
     }
 
     if (mixedRoutes.length > 0) {
       const mixedRoutesFromCache: MixedRoute[] = mixedRoutes.map((cachedRoute) => cachedRoute.route as MixedRoute);
 
+      const beforeGetQuotes = Date.now();
+      metric.putMetric("SwapRouteFromCache_Mixed_GetQuotes_Request", 1, MetricLoggerUnit.Count);
```
mikeki, line 1339: Aren't we missing this metric for V2? The request metric

mikeki (approved):

src/routers/alpha-router/alpha-router.ts
jsy1218: Good catch..

PR #331, opened 2023-08-22
Update HeuristicsFiltering

mikeki:
Bug Fix - Optimization

We are currently iterating several times over the `subgraphPoolsSorted` array, this operation is very slow, consuming most of the time of our v2 quotes.

Refactored the code, we now iterate at most twice over the array, but we will break out of the for loop as soon as we satisfy all our heuristics.

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/functions/get-candidate-pools.ts
+++ b/src/routers/alpha-router/functions/get-candidate-pools.ts
@@ -196,11 +196,9 @@ export async function getV3CandidatePools({
     blockNumber,
     v3PoolSelection: {
       topN,
-      topNDirectSwaps,
       topNTokenInOut,
       topNSecondHop,
       topNSecondHopForTokenAddress,
-      topNWithEachBaseToken,
       topNWithBaseToken,
     },
   } = routingConfig;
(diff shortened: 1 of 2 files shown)


src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -357,133 +300,207 @@ export async function getV3CandidatePools({
...
 
+  const topByTVLUsingTokenInSecondHops = Array.from(topByTVLUsingTokenInSecondHopsMap.values())
+    .reduce((accumulator, currentArray) => accumulator.concat(currentArray), []);
+  const topByTVLUsingTokenOutSecondHops = Array.from(topByTVLUsingTokenOutSecondHopsMap.values())
+    .reduce((accumulator, currentArray) => accumulator.concat(currentArray), []);
+
+  addToAddressSet(topByTVLUsingTokenInSecondHops);
   addToAddressSet(topByTVLUsingTokenOutSecondHops);
 
   const subgraphPools = _([
```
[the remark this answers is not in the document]
mikeki: I think like 498 it first gets mapped to an string, and thats why it works

test/unit/routers/alpha-router/functions/get-candidate-pools.test.ts
```diff
@@ -140,8 +140,10 @@ describe('get candidate pools', () => {
 
     expect(
       mockV3PoolProvider.getPools.calledWithExactly([
-        [DAI, USDC, FeeAmount.LOW],
+        [DAI, USDC, FeeAmount.HIGH],
```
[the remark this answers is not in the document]
mikeki: Previously we would iterate over the whole list of pools to find all the direct swap pools that exist, if one of these pools is missing (like some pairs only have 1 but not the other 3), we would still iterate over the whole list of pools.

This seems to me like a waste of computing cycles, so I just decided to always add the 4 direct swap pools regardless, and that way we might be able to break out of the loop earlier, but that means that now we will always have the 4 direct swap pools

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -357,133 +300,207 @@ export async function getV3CandidatePools({
...
+    if (
+      topByBaseWithTokenIn.length < topNWithBaseToken &&
+      (
+        (baseTokensAddresses.includes(subgraphPool.token0.id) && subgraphPool.token1.id == tokenInAddress) ||
+        (baseTokensAddresses.includes(subgraphPool.token1.id) && subgraphPool.token0.id == tokenInAddress)
+      )
+    ) {
+      topByBaseWithTokenIn.push(subgraphPool);
+      // This pool has been added to the heuristic, continue with the next pool
+      continue;
```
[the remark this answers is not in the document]
mikeki: discussed offline, I think in this case it is easier to follow the continues given the complexity of the heuristics

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -357,149 +278,265 @@ export async function getV3CandidatePools({
...
+      ) &&
+      (
+        routeType === TradeType.EXACT_INPUT && (
+          (subgraphPool.token0.id == wrappedNativeAddress && subgraphPool.token1.id == tokenOutAddress) ||
+          (subgraphPool.token1.id == wrappedNativeAddress && subgraphPool.token0.id == tokenOutAddress)
+        ) ||
+        routeType === TradeType.EXACT_OUTPUT && (
+          (subgraphPool.token0.id == wrappedNativeAddress && subgraphPool.token1.id == tokenInAddress) ||
+          (subgraphPool.token1.id == wrappedNativeAddress && subgraphPool.token0.id == tokenInAddress)
+        )
```
zhongeric, line 371: nit: ik this was from existing code but this conditional is really hard to follow, would support breaking it up into helper functions like `poolContainsNative` etc.

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -357,149 +278,265 @@ export async function getV3CandidatePools({
...
     ...topByTVLUsingTokenOutSecondHops,
-  ])
-    .compact()
-    .uniqBy((pool) => pool.id)
-    .value();
+  ];
+  const subgraphPoolsSet: Set<V3SubgraphPool> = new Set();
+  for (const pool of topPoolsByAllHeuristics) {
+    subgraphPoolsSet.add(pool);
+  }
```
zhongeric, line 531: could we do `[...new Set(topPoolsByAllHeuristics)]`

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -357,149 +278,265 @@ export async function getV3CandidatePools({
     );
   }
 
-  addToAddressSet(top2DirectSwapPool);
+  addToAddressSet(topByDirectSwapPools);
 
+  const baseTokens = baseTokensByChain[chainId] ?? [];
+  const baseTokensAddresses = baseTokens.map((token) => token.address.toLowerCase());
```
jsy1218, line 284: nit: I think it will be _slightly_ quicker if we do lower casing in the if block within the for loop, since topNWithEachBaseToken < baseTokensByChain[mainnet], and topNWithBaseToken < baseTokensByChain[mainnet]

might not impact perf that much, if lower casing is a cheap operation.

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -761,124 +733,225 @@ export async function getV2CandidatePools({
...
+      continue;
+    }
 
-  addToAddressSet(topByTVLUsingTokenIn);
+    // Only consider pools where neither tokens are in the blocked token list.
+    if (blockedTokenListProvider) {
+      const [token0InBlocklist, token1InBlocklist] = await Promise.all([
+        blockedTokenListProvider.getTokenByAddress(subgraphPool.token0.id),
+        blockedTokenListProvider.getTokenByAddress(subgraphPool.token1.id)
+      ]);
```
zhongeric, line 879: hmm i know we're re-iterating over the subgraphPoolsSorted here but is there any way to not have to call the blockedTokenListProvider again since we already do it above?

mikeki: so I thought about this too, and I thought of collecting a list of results, however, the blockedTokenListProvider already contains a caching layer on top, so this is not really a big problem.

src/routers/alpha-router/functions/get-candidate-pools.ts
mikeki: yeah it is definitely hard to follow, I think a lot of these conditions will ahve to change at some point, but my main focus for this branch was to remove the multiple loops.

I can create a helper function, but it is also confusing because all the v2, v3 and mixed candidate pool functions are inside this `get-candidate-pools.ts` file.

I think there's way more room for improvement in here, and I would defer a bit on further changes on this branch until we spend the time refactoring

src/routers/alpha-router/functions/get-candidate-pools.ts
mikeki: I don't think thats necessarily true, it depends of how many iterations we do before we fill `topNWithBaseToken` that could happen in 4 iterations or 1000.

`baseTokenByChain` is usually a super small list, manually curated

src/routers/alpha-router/functions/get-candidate-pools.ts
zhongeric: yeah makes sense

zhongeric (approved):

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -357,149 +278,265 @@ export async function getV3CandidatePools({
...
+      topByTVLUsingTokenOut.length < topNTokenInOut &&
+      (subgraphPool.token0.id == tokenOutAddress || subgraphPool.token1.id == tokenOutAddress)
+    ) {
+      topByTVLUsingTokenOut.push(subgraphPool);
+      continue;
+    }
+  }
+
+  // Add the addresses found so far to our address set
+  addToAddressSet(topByBaseWithTokenIn);
```
jsy1218, line 401: I think we might want to check `poolAddressesSoFar` dedup in `addToAddressSet` with current implementation.

jsy1218: Also helps on perf, since we can dedup across `topByBaseWithTokenIn`, ``topByBaseWithTokenOut`... `topByTVL `

jsy1218: nvm, with the current for-loop each if-continue structure, we are guaranteed to have no duplicates.

jsy1218 (approved):

PR #340, opened 2023-08-22
Revert V3 Candidate Pool Selection

mikeki:
This is a backup plan in case the latest deploy doesn't work as we expect, we would revert the v3 heruistic changes to get the latency back to normal levels


jsy1218 (approved):
fast approval, just in case

PR #344, opened 2023-08-23
Fix Heuristics Bases

mikeki:
Bug Fix
TopByBase is not being considered for second hops
consider topbybase as potential second hops
- **Other information**:


src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -988,7 +1139,9 @@ export async function getV2CandidatePools({
     ...topByTVLUsingTokenOut,
     ...topByTVLUsingTokenInSecondHops,
     ...topByTVLUsingTokenOutSecondHops,
-  ];
+  ])
+    .uniqBy((pool) => pool.id)
```
jsy1218, line 1143: yea that was the missing part I think.

jsy1218 (approved):

PR #357, opened 2023-08-24
Add weth wrap / unwrap gas cost to heuristics and add unit tests for gas models

zhongeric:

Adds an optional parameter `additionalGasOverhead` to ProviderConfig parameter of `buildGasModel` which applies a specified overhead to the gas estimate.

We were not previously accounting for the additional gas cost in trading ETH on V2/V3 pools as it must be wrapped/unwrapped into WETH every swap. This is very important for UniswapX as this is making it harder for X orders to beat out classic if our gas estimate is incorrect.

I also went ahead and added some very basic unit tests to the gas models.

I think my local prettier setup might be wrong, causing the large diff... I ran the prettier command but that changed like 30 files lol so maybe another PR.

TODO: 
- implement for V2 routes, that factory is set up differently than the v3 one and thus we unfortunately don't have access to the original amount/quote tokens, only the wrapped versions, which won't work for detecting if the trade starts or ends in a native token. I think we should address this in a fast follow / or in a larger refactor of the gas heuristic logic.
- change the gas overhead values for chains that don't have canonical WETH9 deployment, but there shouldn't be a huge difference

### DIFF
--- a/src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
+++ b/src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
@@ -8,6 +8,7 @@ import {
   SwapType,
   WRAPPED_NATIVE_CURRENCY,
 } from '../../../..';
+import { ProviderConfig } from '../../../../providers/provider';
 import {
   ArbitrumGasData,
   OptimismGasData,
@@ -26,7 +27,6 @@ import {
   IOnChainGasModelFactory,
 } from '../gas-model';
 
-import { ProviderConfig } from '../../../../providers/provider';
 import {
   BASE_SWAP_COST,
   COST_PER_HOP,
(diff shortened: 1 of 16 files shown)


mikeki (approved):
there's a lint issue with importing the whole `ethers` dependency, we should import a specific module

src/routers/alpha-router/gas-models/v3/v3-heuristic-gas-model.ts
```diff
@@ -379,7 +398,12 @@ export class V3HeuristicGasModelFactory extends IOnChainGasModelFactory {
       .add(hopsGasUse)
       .add(tokenOverhead)
       .add(tickGasUse)
-      .add(uninitializedTickGasUse);
+      .add(uninitializedTickGasUse)
+      .add(
+        providerConfig?.additionalGasOverhead
+          ? providerConfig.additionalGasOverhead
+          : BigNumber.from(0)
```
jsy1218, line 405: `providerConfig?.additionalGasOverhead ?? 0`

marktoda (approved):

jsy1218 (approved):

PR #358, opened 2023-08-24
Get All Tokens From List

passabilities:
	- feature

	- `CachingTokenListProvider.getTokens` only able to fetch specific tokens by addresses

	- if no address array is passed in to `getTokens` it will return a `TokenAccessor` containing **_all_** of the chain's tokens from the token list

- **Other information**:
	- This allows visibility into all of the chain's token from the token list which can be iterated over.

### DIFF
--- a/src/providers/caching-token-list-provider.ts
+++ b/src/providers/caching-token-list-provider.ts
@@ -139,21 +139,36 @@ export class CachingTokenListProvider
     return tokenProvider;
   }
 
-  public async getTokens(_addresses: string[]): Promise<TokenAccessor> {
+  /**
+   * If no addresses array is specified, all tokens in the token list are
+   * returned.
+   *
+   * @param _addresses (optional) The token addresses to get.
+   * @returns Promise<TokenAccessor> A token accessor with methods for accessing the tokens.
+   */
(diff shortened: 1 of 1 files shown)


mikeki:
Your branch will unfortunately conflict with mine: #361, and I dont think we should make these values public, but rather create `getter` methods for the data that you need.

src/providers/caching-token-list-provider.ts
```diff
@@ -139,21 +139,36 @@ export class CachingTokenListProvider
...
     const symbolToToken: Map<string, Token> = new Map();
 
-    for (const address of _addresses) {
-      const token = await this.getTokenByAddress(address);
-      if (!token) {
-        continue;
+    const addToken = (token?: Token) => {
+      if (!token) return;
+      addressToToken.set(token.address.toLowerCase(), token);
+      if (!!token.symbol) {
```
mikeki, line 156: nit: can we do `token.symbol !== undefined` rather than double negation?

mikeki (approved):

mikeki:
@passabilities This was released on @uniswap/smart-order-router@3.16.13

PR #366, opened 2023-08-25
add hits cached routes boolean flag

jsy1218:
Telemetry

We don't have a top-level request latencies breakdown between cached routes vs non-cached routes.

We want this visibility. The only way to know is that [smart-order-router](https://github.com/Uniswap/smart-order-router) starts returning that boolean flag.

- **Other information**:
The boolean needs to be optional, so that the response schema is backward compatible. It's a debug only flag anyway.

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1213,7 +1213,7 @@ export class AlphaRouter
       trade,
       methodParameters,
       blockNumber: BigNumber.from(await blockNumber),
-      hitsCachedRoute: swapRouteFromChain === null,
+      hitsCachedRoute: !swapRouteFromChain && swapRouteFromCache !== null ,
     };
 
     if (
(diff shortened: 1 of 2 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1213,6 +1213,7 @@ export class AlphaRouter
       trade,
       methodParameters,
       blockNumber: BigNumber.from(await blockNumber),
+      hitsCachedRoute: swapRouteFromChain === null,
```
mikeki, line 1216: `!swapRouteFromChain && swapRouteFromCache` ?

jsy1218: Sure. They seems mutually exclusive, but can be safer this way.

jsy1218: It has to be `!swapRouteFromChain && swapRouteFromCache !== null` to be type conformant.

mikeki (approved):

PR #378, opened 2023-08-29
Keep track of tokens visited and avoid visiting back a token

mikeki:
BugFix/Feature

Currently we will compute routes that represent arbitrage opportunities, when we go back to a previously seen token in the same route, this is a problem because we don't operate at the MEV level, so offering these routes to our users is not ideal, nor a good idea.

In this new behavior we keep track of the tokens visited per route, and we avoid visiting it again.

- **Other information**:

### DIFF
--- a/test/unit/routers/alpha-router/functions/compute-all-routes.test.ts
+++ b/test/unit/routers/alpha-router/functions/compute-all-routes.test.ts
@@ -9,8 +9,9 @@ import {
   WRAPPED_NATIVE_CURRENCY,
 } from '../../../../../src';
 import {
+  computeAllMixedRoutes,
   computeAllV2Routes,
-  computeAllV3Routes,
+  computeAllV3Routes
 } from '../../../../../src/routers/alpha-router/functions/compute-all-routes';
 import {
   DAI_USDT,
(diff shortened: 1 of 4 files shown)


test/unit/routers/alpha-router/functions/compute-all-routes.test.ts
```diff
@@ -83,6 +101,17 @@ describe('compute all v2 routes', () => {
     expect(routes).toHaveLength(1);
   });
 
+  test('succeeds to compute all routes with 5 hops. ignoring arbitrage opportunities', async () => {
+    const pools = [DAI_USDT, USDC_DAI, USDC_USDT, USDC_WETH, WETH_USDT, WBTC_WETH];
+    const routes = computeAllV2Routes(USDC, WRAPPED_NATIVE_CURRENCY[1]!, pools, 5);
+
+    routes.forEach((route) => {
+      expect(route.pairs).not.toEqual([USDC_USDT, DAI_USDT, USDC_DAI, USDC_WETH]);
+      expect(route.pairs).not.toEqual([USDC_DAI, DAI_USDT, USDC_USDT, USDC_WETH]);
```
zhongeric, line 110: can you add a test for mixed routes that it prevents something like

V2 (USDC -> DAI)  -> V3 (DAI -> USDC) -> V3 (USDC-> WETH) ?

I think i've seen that happen before

zhongeric (approved):
lgtm just request to cover mixed route case

jsy1218:
One alternative is instead of using new set to keep track of used token, we can introduce a new data structure `UsedToken` with `visited` boolean flag to indicate whether it's used.

This is just in case that the set becomes memory intensive. After seeing how lodash code removals helped on memories and overall perf, a bit cautious on getting bit.

PR #380, opened 2023-08-30
Only quote the 100% amount when there's only 1 route

mikeki:
small optimization

currently we quote all the distribution percents even if there's only 1 route

we will only quote the 100% amount when we have only 1 route

- **Other information**:


jsy1218 (approved):

jsy1218:
Might be a good idea to leave a comment in https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/config.ts, that in case of single route, the `distributionPercent` config from sor will not be honored, and instead there's a simple heuristics to use a single route for on-chain quoting.

PR #394, opened 2023-09-05
Feature: add token properties provider

jsy1218:
Feature

Currently we don't have a way to know the token tax for the fee-on-transfer tokens

We will fetch the tax/fee bps for fee-on-transfer tokens. Since we don't know which tokens will be fee-on-transfer tokens for routing, we will have to add a caching layer to reduce the number of infura calls.

- **Other information**:
Tested locally via CLI, with some additional console logs
Screenshot 2023-09-06 at 9 21 47 PM

0xab98093c7232e98a47d7270ce0c1c2106f61c73b (BLASTERS (BLAST)) has 3% buy tax and 3.4% sell tax.
0x8ef32a03784c8fd63bbf027251b9620865bd54b6 (Bullet Game Betting Token (BULLET)) has 5% buy tax and 5% sell tax.


src/providers/token-fee-fetcher.ts
```diff
@@ -0,0 +1,103 @@
...
+type Address = string;
+
+export type TokenFeeResult = {
+  buyFeeBps: BigNumber;
+  sellFeeBps: BigNumber;
+};
+export type TokenFeeMap = Record<Address, TokenFeeResult>;
+
+// address at which the FeeDetector lens is deployed
+const FEE_DETECTOR_ADDRESS = '0x57eC54d113719dDE9A90E6bE807524a86560E89D';
```
mikeki, line 19: We'll need this per chain eventually, it seems like we will start adding V2 support soon-ish per governance vote.

src/providers/token-fee-fetcher.ts
```diff
@@ -0,0 +1,103 @@
...
+      >({
+        address: this.tokenFeeAddress,
+        contractInterface: TokenFeeDetector__factory.createInterface(),
+        functionName: 'validate',
+        functionParams: functionParams,
+        providerConfig,
+        additionalConfig: {
+          gasLimitPerCallOverride: this.gasLimitPerCall,
+        },
+      });
```
mikeki, line 74: I wonder if we it would be faster to call the contract directly instead of through a multicall?

I've been having the theory that the multicall contracts are actually slower

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+export interface ITokenPropertiesProvider {
+  getTokensProperties(
+    tokens: Token[],
+    providerConfig?: ProviderConfig
+  ): Promise<TokenPropertiesMap>;
+}
+
+export class TokenPropertiesProvider implements ITokenPropertiesProvider {
+  private CACHE_KEY = (chainId: ChainId, address: string) =>
+    `token-${chainId}-${address}`;
```
mikeki, line 39: maybe `token-properties-${chainId}-${address}`?

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+        addressesToFetchFeesOnchain,
+        providerConfig
+      );
+
+      for (const address of addressesToFetchFeesOnchain) {
+        const tokenFee = tokenFeeMap[address];
+        if (tokenFee) {
+          await this.tokenPropertiesCache.set(
+            this.CACHE_KEY(this.chainId, address),
+            { tokenFeeResult: tokenFee }
```
mikeki, line 89: should we include the blockNumber when the fee was last seen? Do we even care about that or just cache until it ttls?

jsy1218: I think we want to have blockSeen as part of the `DynamoTokenPropertiesProvider` implementation, where after getting the entry, we will need to see if the blockSeen is beyond hardcoded blocksToLive.

src/providers/token-fee-fetcher.ts
jsy1218: got it, we already pass chain id into constructor. We can make FEE_DETECTOR_ADDRESS a map of `ChainId to deployer address

src/providers/token-fee-fetcher.ts
jsy1218: Seems like so. @andysmith415 also had the same conclusion from fast poc. Can make them direct contract calls.

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+    tokens: Token[],
+    providerConfig?: ProviderConfig
+  ): Promise<TokenPropertiesMap> {
+    const tokenToResult: TokenPropertiesMap = {};
+    const addressesToFetchFeesOnchain: string[] = [];
+    const addressesRaw = this.buildAddressesRaw(tokens);
+
+    // Check if we have cached token validation results for any tokens.
+    for (const address of addressesRaw) {
+      const cachedValue = await this.tokenPropertiesCache.get(
```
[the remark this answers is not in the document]
jsy1218: Good idea, we can support batch get from `ICache`, and dynamo has the native batch get

src/providers/token-fee-fetcher.ts
jsy1218: I used the direct contract call, and the code is also looking a lot cleaner than to use multicall

src/providers/token-properties-provider.ts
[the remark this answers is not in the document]
jsy1218: Added `batchGet(keys: Set<string>): Promise<Record<string, T | undefined>>` and `NodeJSCache` implementation of it. Then in the dynamo caching implementation, I think we can use `batchGet` to save additional network trips.

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+    );
+
+    if (addressesToFetchFeesOnchain.length > 0) {
+      const tokenFeeMap = await this.tokenFeeFetcher.fetchFees(
+        addressesToFetchFeesOnchain,
+        providerConfig
+      );
+
+      for (const address of addressesToFetchFeesOnchain) {
+        const tokenFee = tokenFeeMap[address];
```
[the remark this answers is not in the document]
jsy1218: I was thinking to parallelize calling `TokenValidatorProvider` and `TokenPropertiesProvider` to save the latencies. But yes, then `TokenPropertiesProvider` will always go out for non-FOT tokens, and doesn't seem to worth it. Since `TokenValidatorProvider` has a in-memory cache, I think calling `TokenValidatorProvider` within `TokenPropertiesProvider` makes sense.

src/providers/token-fee-fetcher.ts
```diff
@@ -0,0 +1,97 @@
...
+    const functionParams = addresses.map((address) => [
+      address,
+      this.BASE_TOKEN,
+      this.amountToFlashBorrow,
+    ]) as [string, string, string][];
+
+    const results = await Promise.all(functionParams.map(async ([address, baseToken, amountToBorrow]) => {
+      log.info(
+        { address, baseToken, amountToBorrow },
+        `Called validate on-chain for token ${address}`
```
cgkol, line 77: is this left here by design? thats gonna be a loooot of logs

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (cachedValue) {
+        tokenToResult[address] = cachedValue;
+      } else if (this.allowList.has(address)) {
+        tokenToResult[address] = DEFAULT_TOKEN_PROPERTIES_RESULT;
+      } else {
+        addressesToFetchFeesOnchain.push(address);
+      }
+    }
+
+    log.info(
```
cgkol, line 70: again probably too many logs

src/providers/token-properties-provider.ts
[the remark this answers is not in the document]
cgkol: @andysmith415 can you help me understand how negative cache would speed up things? I am actually unfamiliar since i've never used it. How is it different than realizing that something is empty? You just dont have to go through the array one more time to see whats left?

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (addressesToFetchFeesOnchain.length > 0) {
+      const tokenFeeMap = await this.tokenFeeFetcher.fetchFees(
+        addressesToFetchFeesOnchain,
+        providerConfig
+      );
+
+      for (const address of addressesToFetchFeesOnchain) {
+        const tokenFee = tokenFeeMap[address];
+        if (tokenFee) {
+          await this.tokenPropertiesCache.set(
```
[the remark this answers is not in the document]
cgkol: hmm good callout! Then again, given how rare this happens might be an overkill? Whats the TTL gonna be 1 day?

cgkol (commented):
I think this is missing a few tests to check the provider properties logic right?

src/providers/token-fee-fetcher.ts
jsy1218: We have 10% log sampling in routing-api. Unsure about client-side logging though. Can remove.

jsy1218:
> I think this is missing a few tests to check the provider properties logic right?

Working on the unit test coverage as follow-up, and then will convert the PR from draft to ready for review.

src/providers/token-properties-provider.ts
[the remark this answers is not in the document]
jsy1218: I think what Andy means here is that before we fetch the token fees, we can have a cache to remember tokens that are not FOTs. Since `TokenValidatorProvider` exactly fits the negative cache purpose here (it contains the in-memory cache for FOT and non-FOT token results), it makes sense to call `TokenValidatorProvider` to check whether the token is FOT, before calling out to `TokenFeeFetcher` to fetch the token fees, even for non-FOT tokens.

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+    const addressesRaw = this.buildAddressesRaw(tokens);
+
+    const tokenProperties = await this.tokenPropertiesCache.batchGet(addressesRaw)
+
+    // Check if we have cached token validation results for any tokens.
+    for (const address of addressesRaw) {
+      const cachedValue = tokenProperties[address];
+      if (cachedValue) {
+        tokenToResult[address] = cachedValue;
+      } else if (this.allowList.has(address)) {
```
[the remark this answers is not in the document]
jsy1218: I made it such that `TokenPropertiesProvider` calls out to `TokenValidatorProvider` first. As part of it, if the token is in the allowlist, we hardcode that token to UNKNOWN result type, so that the allowlist tokens will not call the `TokenFeeFetcher` since they are not FOT.

src/providers/token-properties-provider.ts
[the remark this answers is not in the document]
jsy1218: Yes, we plan TTL to be 1 day. This is nice optimization. I think the main thing is we got bite by the router latencies so much, so we want to apply as much due diligence as possible going forward, right?

src/providers/token-properties-provider.ts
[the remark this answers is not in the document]
jsy1218: I updated the `TokenPropertiesProvider` to ensure the negative cache `TokenValidatorProvider` blocks `TokenPropertiesProvider` from querying `TokenFeeFetcher` for non-FOT tokens. Also added the unit test coverage to cover this exact negative cache case.

src/providers/token-fee-fetcher.ts
```diff
@@ -0,0 +1,117 @@
...
+  sellFeeBps?: BigNumber;
+};
+export type TokenFeeMap = Record<Address, TokenFeeResult>;
+
+// address at which the FeeDetector lens is deployed
+const FEE_DETECTOR_ADDRESS = (chainId: ChainId) => {
+  switch (chainId) {
+    case ChainId.MAINNET:
+    default:
+      return '0x57eC54d113719dDE9A90E6bE807524a86560E89D';
```
jsy1218, line 33: Just noticed that interface uses 0x19C97dc2a25845C7f9d1d519c8C2d4809c58b43f. Need to follow-up with protocol folks on which contract is the right one.

jsy1218: Most likely 0x19C97dc2a25845C7f9d1d519c8C2d4809c58b43f. 0x19C97dc2a25845C7f9d1d519c8C2d4809c58b43f don't have the `UniswapV2: Locked` error.

cgkol (approved):
✅ awesome Siyu! Great test coverage too. I would also wait for Miguel or Andy to bless it tho

src/providers/token-properties-provider.ts
```diff
@@ -0,0 +1,152 @@
...
+                tokenFeeResult: tokenFee,
+                tokenValidationResult: TokenValidationResult.FOT,
+              }
+            );
+          } else {
+            return Promise.resolve(true);
+          }
+        })
+      );
+    }
```
mikeki, line 135: Seems like we could use `batchWrite` here: https://github.com/Uniswap/routing-api/blob/88c83324a549be81b4f8bcfdf3adaadc1bcf75c8/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L416

src/routers/alpha-router/alpha-router.ts
```diff
@@ -33,12 +33,12 @@ import {
   Simulator,
   StaticV2SubgraphProvider,
   StaticV3SubgraphProvider,
-  SwapRouterProvider,
+  SwapRouterProvider, TokenPropertiesProvider,
```
mikeki, line 36: nit: prettier

src/providers/token-properties-provider.ts
jsy1218: I think I will have this as a separate PR follow-up.

najy1366:
What is the role of  Ryoshi token here?

src/providers/token-properties-provider.ts
najy1366: > Seems like we could use `batchWrite` here: https://github.com/Uniswap/routing-api/blob/88c83324a549be81b4f8bcfdf3adaadc1bcf75c8/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L416

What is the role of Ryoshi token here

najy1366:
Thankful
I read the content but it is complicated.
Please explain the role of Rioshi in this content in simple language
Thankful

PR #399, opened 2023-09-14
Fix: quote provider has the enable fot flag awareness

jsy1218:
Big fix

There's a bug in the FOT quote and routes calculation path, so that it would cause interface to double quote FOT tax immediately, if we ship https://github.com/Uniswap/routing-api/pull/362 today.

Let's use BULLET => WETH as an example. Hypothetically interface will issue two requests:

1. BULLET => WETH exact in with enableFeeOnTransferFeeFetching = true. Then it will writes the routes DB with the cached routes with the pool reserves with the buyFeeBps and sellFeeBps . Interface issued this request as part of their QA/testing.
2. Degen uses interface issue a BULLET => WETH exact in quote. Since interface has not enabled to pass in enableFeeOnTransferFeeFetching , routing-api will have useCachedRoutes = true, and read from cached routes with the pool reserves with the buyFeeBps and sellFeeBps . Then in sor, it will calculate the routes and quotes with the buyFeeBps and sellFeeBps .

The problem is in (2), interface has no awareness that the quote and routes are from cached routes, and they have the hardcoded FOT tax logic in prod today https://github.com/Uniswap/interface/blob/main/src/hooks/useSwapTaxes.ts. Since [smart-order-router](https://github.com/Uniswap/smart-order-router) also quoted with FOT tax in [quote-provider](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v2/quote-provider.ts#L74-L110), this is effectively double FOT taxation in prod, before we enable FOT backend support to 100%. And since cached routes hit rate is over +80%, this bug is considered a blocker before we ship https://github.com/Uniswap/routing-api/pull/362.

Quote provider doesn't calculate FOT tax. As such because it also doesn't pass the FOT tax to the `outputAmount`/`inputAmount`, the routes returned from routing-api will not include the FOT tax in both token construct as well as the `outputAmount`/`inputAmount`.

- **Other information**:

### DIFF
--- a/src/providers/v2/quote-provider.ts
+++ b/src/providers/v2/quote-provider.ts
@@ -51,7 +51,12 @@ export class V2QuoteProvider implements IV2QuoteProvider {
     routes: V2Route[],
     providerConfig: ProviderConfig
   ): Promise<{ routesWithQuotes: V2RouteWithQuotes[] }> {
-    return this.getQuotes(amountIns, routes, TradeType.EXACT_INPUT, providerConfig);
+    return this.getQuotes(
+      amountIns,
+      routes,
+      TradeType.EXACT_INPUT,
+      providerConfig
+    );
   }
 
   public async getQuotesManyExactOut(
@@ -59,14 +64,19 @@ export class V2QuoteProvider implements IV2QuoteProvider {
     routes: V2Route[],
     providerConfig: ProviderConfig
   ): Promise<{ routesWithQuotes: V2RouteWithQuotes[] }> {
-    return this.getQuotes(amountOuts, routes, TradeType.EXACT_OUTPUT, providerConfig);
+    return this.getQuotes(
+      amountOuts,
+      routes,
+      TradeType.EXACT_OUTPUT,
+      providerConfig
+    );
   }
 
   private async getQuotes(
     amounts: CurrencyAmount[],
     routes: V2Route[],
     tradeType: TradeType,
-    providerConfig: ProviderConfig,
+    providerConfig: ProviderConfig
   ): Promise<{ routesWithQuotes: V2RouteWithQuotes[] }> {
     const routesWithQuotes: V2RouteWithQuotes[] = [];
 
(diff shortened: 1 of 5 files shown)


src/providers/v2/quote-provider.ts
```diff
@@ -75,14 +82,32 @@ export class V2QuoteProvider implements IV2QuoteProvider {
...
+                  const [outputAmountNew] = pair.getOutputAmount(outputAmountWithSellFeeBps);
+                  outputAmount = outputAmountNew;
+                } else if (pair.token1.equals(outputAmount.currency) && pair.token1.sellFeeBps?.gt(BigNumber.from(0))) {
+                  const outputAmountWithSellFeeBps = CurrencyAmount.fromRawAmount(pair.token1, outputAmount.quotient);
+                  const [outputAmountNew] = pair.getOutputAmount(outputAmountWithSellFeeBps);
+                  outputAmount = outputAmountNew;
+                } else {
+                  const [outputAmountNew] = pair.getOutputAmount(outputAmount);
+                  outputAmount = outputAmountNew;
+                }
```
mikeki, line 110: Did something change in this block of code?

jsy1218: I realized a buggy scenario, which will be common in prod, if we ship routing-api PR today. bcc from the descriptions:

_There's a bug in the FOT quote and routes calculation path, so that it would cause interface to double quote FOT tax immediately, if we ship https://github.com/Uniswap/routing-api/pull/362 today._

_Let's use BULLET => WETH as an example. Hypothetically interface will issue two requests:_

_BULLET => WETH exact in with enableFeeOnTransferFeeFetching = true. Then it will writes the routes DB with the cached routes with the pool reserves with the buyFeeBps and sellFeeBps . Interface issued this request as part of their QA/testing._
_Degen uses interface issue a BULLET => WETH exact in quote. Since interface has not enabled to pass in enableFeeOnTransferFeeFetching , routing-api will have useCachedRoutes = true, and read from cached routes with the pool reserves with the buyFeeBps and sellFeeBps . Then in sor, it will calculate the routes and quotes with the buyFeeBps and sellFeeBps ._
_The problem is in (2), interface has no awareness that the quote and routes are from cached routes, and they have the hardcoded FOT tax logic in prod today https://github.com/Uniswap/interface/blob/main/src/hooks/useSwapTaxes.ts. Since [smart-order-router](https://github.com/Uniswap/smart-order-router) also quoted with FOT tax in [quote-provider](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v2/quote-provider.ts#L74-L110), this is effectively double FOT taxation in prod, before we enable FOT backend support to 100%. And since cached routes hit rate is over +80%, this bug is considered a blocker before we ship https://github.com/Uniswap/routing-api/pull/362._

mikeki: oh got it got it previously we were gating this on the presence of sellFeeBps instead of the flag 👍

mikeki (approved):

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1178,42 +1178,12 @@ export class AlphaRouter
       cacheMode !== CacheMode.Darkmode &&
       swapRouteFromChain
     ) {
-      const tokenPropertiesMap = await this.tokenPropertiesProvider.getTokensProperties([tokenIn, tokenOut], providerConfig);
```
jsy1218, line 1181: FYI, We also no longer need to call `tokenPropertiesProvider.getTokensProperties` for `cachedRoutes.tokenIn` and `cachedRoutes.tokenOut`, because `swapRouteFromChain.routes` will contain the pool reserve with FOT tax, and can apply to `QuoteProvider.getQuotes` correctly. (i.e.,  `cachedRoutes.tokenIn` and `cachedRoutes.tokenOut` FOT tax will get discarded, and Pair.reserve FOT tax will be reapplied in `QuoteProvider.getQuotes`

PR #401, opened 2023-09-14
add tenderly env variables to github workflow

jsy1218:
Bug fix

We forgot to pass in the tenderly env variables to github workflow, so that the integ-test doesn't run with tenderly simulation.

Github workflow integ-test runs with tenderly simulation now.

- **Other information**:


test/integ/routers/alpha-router/alpha-router.integration.test.ts
```diff
@@ -3039,7 +3043,9 @@ describe('quote for other networks', () => {
 
             it(`${native} -> erc20`, async () => {
               const tokenIn = nativeOnChain(chain);
-              const tokenOut = erc2;
+              // TODO ROUTE-64: Remove this once smart-order-router supports ETH native currency on BASE
+              // see https://uniswapteam.slack.com/archives/C021SU4PMR7/p1691593679108459?thread_ts=1691532336.742419&cid=C021SU4PMR7
+              const tokenOut = chain == ChainId.BASE ? USDC_ON(ChainId.BASE) : erc2
```
zhongeric, line 3048: isn't this testing ETH input so you don't have to special case the output token?

zhongeric (approved):

test/integ/routers/alpha-router/alpha-router.integration.test.ts
jsy1218: Sadly, routing team missed to implement `BaseNativeCurrency` in https://github.com/Uniswap/smart-order-router/blob/main/src/util/chains.ts#L575-L592, so that when tokenIn&tokenOut pair is ETH&WETH, sor will auto wrap ETH to WETH, and we hit invariant error. 

For now, we have a ticket to follow up ROUTE-64

PR #405, opened 2023-09-15
Fix: integ-test mixed route longer timeout to avoid flaky test failures

jsy1218:
Bug fix

Sometimes mixed route BOND -> APE test can fail due to timeout

Double BOND -> APE test timeout

- **Other information**:


jsy1218:
Close to fix all integ-test issue in https://github.com/Uniswap/smart-order-router/pull/401

PR #406, opened 2023-09-22
logging more around fot tax

jsy1218:
Observability

We see sporadic FOT quote without FOT tax. We have a theory, but we need to improve logging and metrics to be sure.

Add more logging and metrics.

Those new metrics together will tell us that when the FOT tax is not returned for FOT token, it will tell us:

1. if it's due to the token-validator contract call failure, then `TokenValidatorProviderValidateFailed` will increment and shows up in the log insights within the same request id. Also the logs containining 'Failed to validate token' will definitively show up now, vs at 10% chance previously.
2. if token-validator contract call succeeds, but somehow contract returns incorrect validation result, then `TokenValidatorProviderValidateSuccess` will show up, and `TokenValidatorProviderValidateCacheMissResult0` will show up as well. Note only `TokenValidatorProviderValidateCacheMissResult1` indicates it's FOT token.
3. If token-validator has cache hit, and didn't call contract, but cache value was not FOT by mistake, then `TokenValidatorProviderValidateCacheHitResult0` shows up, but none of `TokenPropertiesProviderCacheHitTokenFeeResultExists${feeBps}` & `TokenPropertiesProviderCacheHitTokenFeeResultNotExists` & `TokenPropertiesProviderTokenFeeResultCacheMissExists${feeBps}` & `TokenPropertiesProviderTokenFeeResultCacheMissNotExists`. Note only `TokenValidatorProviderValidateCacheHitResult1` indicates it's FOT token.
4. If token-fee-fetcher contract call fails, then `TokenFeeFetcherFetchFeesFailure` will show up.
5. If token-fee-fetcher contract call succeeds, but it doesn't return any fee, then `TokenPropertiesProviderTokenFeeResultCacheMissNotExists` will show up.
6. If token-fee-fetcher contract call succeeds, but somehow incorrectly returns 0 bps for FOT tax, then `TokenPropertiesProviderTokenFeeResultCacheMissExists0` will show up.
7. If token-fee-fetcher has cache hit, and didn't call contract, but cache value does not contain FOT tax by mistake, then `TokenPropertiesProviderCacheHitTokenFeeResultNotExists` will show up.

We think 1 is what's happening in beta, but we can use those metrics 

- **Other information**:
Below is a screenshot of AWS log insights having successful token validator contract call and token fee contract call, with both cache misses:
Screenshot 2023-09-22 at 11 08 51 AM

### DIFF
--- a/src/providers/token-properties-provider.ts
+++ b/src/providers/token-properties-provider.ts
@@ -91,6 +91,14 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
       const cachedValue = tokenProperties[address];
       if (cachedValue) {
         metric.putMetric("TokenPropertiesProviderBatchGetCacheHit", 1, MetricLoggerUnit.Count)
+        const tokenFee = cachedValue.tokenFeeResult;
+        const tokenFeeResultExists: BigNumber | undefined = tokenFee && (tokenFee.buyFeeBps || tokenFee.sellFeeBps)
+
+        if (tokenFeeResultExists) {
+          metric.putMetric(`TokenPropertiesProviderCacheHitTokenFeeResultExists${tokenFeeResultExists}`, 1, MetricLoggerUnit.Count)
+        } else {
+          metric.putMetric(`TokenPropertiesProviderCacheHitTokenFeeResultNotExists`, 1, MetricLoggerUnit.Count)
+        }
 
         tokenToResult[address] = cachedValue;
       } else if (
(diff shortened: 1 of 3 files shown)


src/providers/token-properties-provider.ts
```diff
@@ -89,6 +90,16 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
     for (const address of addressesRaw) {
       const cachedValue = tokenProperties[address];
       if (cachedValue) {
+        metric.putMetric("TokenPropertiesProviderBatchGetCacheHit", 1, MetricLoggerUnit.Count)
+        const tokenFee = cachedValue.tokenFeeResult;
+        const tokenFeeResultExists: BigNumber | undefined = tokenFee && (tokenFee.buyFeeBps || tokenFee.sellFeeBps)
+
+        if (tokenFeeResultExists) {
+          metric.putMetric(`TokenPropertiesProviderCacheHitTokenFeeResultExists${tokenFeeResultExists}`, 1, MetricLoggerUnit.Count)
```
mikeki, line 98: do you know how this metric will look like in practice?

mikeki (approved):

src/providers/token-properties-provider.ts
jsy1218: Yep, I'm judging from `TokenPropertiesProviderCacheMissTokenFeeResultExists` from the screenshot, it will look like `TokenPropertiesProviderCacheHitTokenFeeResultExists0`, `TokenPropertiesProviderCacheHitTokenFeeResultExists300`, etc. The trick here as long as total possibility of feeBps is not crazily high, we won't hit metrics cardinalities under the single namespace in routing-api.

PR #408, opened 2023-09-25
Fix: Remove token validator provider calls from token properties provider

jsy1218:
Bug fix

Token validator providers can return out-of-order results for the input token addresses. Coupled with the long caching TTL across all token properties entries, it can provider very negative experiences to the clients.

Token properties provider remove the calls to token validator providers. Also enhanced the `NodeJSCache` to accept per-entry TTL, so that token properties provider can set differentiated per-entry TTL depending on if it's positive cache or negative cache.

- **Other information**:
tested on the routing-api integ-test side via personal AWS account.

### DIFF
--- a/src/providers/token-properties-provider.ts
+++ b/src/providers/token-properties-provider.ts
@@ -20,6 +20,8 @@ import {
 export const DEFAULT_TOKEN_PROPERTIES_RESULT: TokenPropertiesResult = {
   tokenFeeResult: DEFAULT_TOKEN_FEE_RESULT,
 };
+export const POSITIVE_CACHE_ENTRY_TTL = 600; // 10 minutes in seconds
+export const NEGATIVE_CACHE_ENTRY_TTL = 10; // 10 seconds
 
 type Address = string;
 export type TokenPropertiesResult = {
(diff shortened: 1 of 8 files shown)


src/providers/token-properties-provider.ts
```diff
@@ -11,14 +13,15 @@ import {
...
   TokenValidationResult,
 } from './token-validator-provider';
-import { BigNumber } from '@ethersproject/bignumber';
+
 
 export const DEFAULT_TOKEN_PROPERTIES_RESULT: TokenPropertiesResult = {
   tokenFeeResult: DEFAULT_TOKEN_FEE_RESULT,
 };
+export const POSITIVE_CACHE_ENTRY_TTL = 600; // 10 minutes in seconds
+export const NEGATIVE_CACHE_ENTRY_TTL = 10; // 10 seconds
```
mikeki, line 24: Just curioous why is the negative cache ttl so small?

jsy1218: It's a heuristic based on the mobile UX feedback. When we mistakenly cache the FOT token with no fees, on the clients side, it shows the UI doesn't populate the FOT taxes. This seems like a pretty bad UX. By making it shorter, we can make the router refresh and return the FOT tax sooner, just in case if we cache the FOT token with no fees for some reason, e.g. bug, failed contract calls, etc.

Since the token properties provider constructor accepts both positive and negative TTLs, it's easy to adjust on the routing-api side, as we roll out and learn more UX.

mikeki: I think the concern is that the majority of tokens will be in the set of "negative" cache, and if we ttl every 10 seconds, then thats a lot of rpc calls.

Could we set this to 10 minutes too and see how bad it is, before we go with doing it 10 seconds from the get go?

jsy1218: sure

mikeki (approved):

PR #395, opened 2023-09-07
Feature: V2 routes quote with fee-on-transfer fees off the quote amount

jsy1218:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature

- **What is the current behavior?** (You can also link to an open issue here)
V2 quotes do not consider FOT fees, resulting in higher than usual fees

- **What is the new behavior (if this is a feature change)?**
V2 quotes will consider FOT fees. When we get the eligible pools as part of v2 route discovery, we concurrently get the FOT fees, in addition to reserves.
Also since web already detect FOT tax and shave off quote of the tax rate, we need to expose a feature flag so that web doesn't double tax against the quote.

- **Other information**:
amountOut = B * 997 * amountIn * (1 - sellFeeBps / 10000) / (1000 * A + 997 * amountIn) * (1 - sellFeeBps / 10000) 
amountIn = A * amountOut / (1 - buyFeeBps / 10000) * 1000 / (B - amountOut / (1 - buyFeeBps / 10000)) / 997

tested in following cases:

- [x]  BLAST => BULLET exact in, both are FOT tokens
./bin/cli quote --tokenIn 0xab98093c7232e98a47d7270ce0c1c2106f61c73b  --tokenOut 0x8ef32a03784c8fd63bbf027251b9620865bd54b6  --amount 100000 --exactIn --protocols v2,v3,mixed
https://app.warp.dev/block/N6HaxYjmubtDZzr9fH6aTR

Route path:
BLAST => WETH => BULLET

First reserves pair quote:
BLAST => WETH
WETH amountOut = 4.113888521182481518 * 997 * 100000 * (1 - 0.03) / (1000 * 137784644.932841276+ 997 * 100000 * (1 - 0.03))
= 0.002885452119862558

Second reserves pair quote:
WETH => BULLET
BULLET amountOut = 4845053.62651238 * 997 * 0.002885452119862558 * (1 - 0.05) / (1000 * 42.398745781910187318 + 997 * 0.002885452119862558 * (1 - 0.05))
= 312.28329139

prod interface quote amount:
Screenshot 2023-09-10 at 10 20 05 PM

Amount close enough to prod. because interface uses total tax off the output quote.

local sor quote amount:
Best Route:
```
[V2] 100.00% = BLAST -- [0x6aDB403912608FffFd8Fe623a7DB35f59e4b4F43] --> WETH -- [0x0D0A1767da735F725f41c4315E072c63Dbc6ab3D] --> BULLET
[...]

PR #419, opened 2023-09-28
fix: add eth wrap unwrap overhead for v2 routes

zhongeric:
Realized that I put this for TODO but forgot to get back around to it.

It's not super elegant since the way we request for v2 quotes is a bit different than v3, as the gas model doesn't get created for each route/amount combination, but rather at the very beginning. However, we can assume that all input `amounts` are of the same token and that there is at least one element in the array so we can make it work.

### DIFF
--- a/src/routers/alpha-router/quoters/v2-quoter.ts
+++ b/src/routers/alpha-router/quoters/v2-quoter.ts
@@ -32,6 +32,7 @@ import { IGasModel, IV2GasModelFactory } from '../gas-models';
 import { BaseQuoter } from './base-quoter';
 import { GetQuotesResult } from './model/results/get-quotes-result';
 import { GetRoutesResult } from './model/results/get-routes-result';
+import { NATIVE_OVERHEAD } from '../gas-models/v3/gas-costs';
 
 export class V2Quoter extends BaseQuoter<V2CandidatePools, V2Route> {
   protected v2SubgraphProvider: IV2SubgraphProvider;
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/quoters/v2-quoter.ts
```diff
@@ -162,8 +170,10 @@ export class V2Quoter extends BaseQuoter<V2CandidatePools, V2Route> {
       gasPriceWei,
       poolProvider: this.v2PoolProvider,
       token: quoteToken,
-      providerConfig: _routingConfig,
-      // TODO: implement wrap overhead for v2 routes
+      providerConfig: {
+        ..._routingConfig,
+        additionalGasOverhead: NATIVE_OVERHEAD(this.chainId, amountToken, quoteToken)
```
mikeki, line 175: I'm a bit confused about what this line does, can you explain a little bit?

Is there any way in which this could mess up something?

jsy1218: I think you can introduce `additionalGasOverhead` within `AlphaRouterConfig`, and that should take care of it here. Because https://github.com/Uniswap/smart-order-router/pull/395/files#diff-2fe4a092f36b8091262f7035638dc00530816382f3a1cca86927a6fd712e9c04R987-R991 made it such that top-level `ProviderConfig` always contains the `additionalGasOverhead`

src/routers/alpha-router/quoters/v2-quoter.ts
```diff
@@ -140,6 +141,13 @@ export class V2Quoter extends BaseQuoter<V2CandidatePools, V2Route> {
     if (gasPriceWei === undefined) {
       throw new Error('GasPriceWei for V2Routes is required to getQuotes');
     }
+    // throw if we have no amounts or if there are different tokens in the amounts
+    if (amounts.length == 0 || !amounts.every((amount) => amount.currency.equals(amounts[0]!.currency))) {
+      throw new Error('Amounts must have at least one amount and must be same token');
```
jsy1218, line 146: ok to validate and throw, although I don't think this error condition will ever happen, due to the implementation in `getAmountDistribution`

src/routers/alpha-router/quoters/v2-quoter.ts
jsy1218: As for the context, it was probably a follow-up to https://github.com/Uniswap/smart-order-router/pull/357. Nice to have this fix, although I don't think it will actually take much effect, since we always enable tenderly simulation for gas estimate. The only scenario I can see this change can help is when there are multiple routes to quote against, and v2 route quote is slightly better than v3 route quote or mixed route quote after adjusting the gas. But this v2 route quote is actually worse, because it didn't take into account the wrap/unwrap overhead. And this probably has to be the ETH/WETH token(s) in the v2 route-through, not token in or token out, that are not in the v3/mixed route.

jsy1218 (approved):

PR #421, opened 2023-09-29
Update SOR to let v2-sdk handle the FOT calculation

mikeki:
# Let V2-SDK handle the calculation for FOT

There is a bug in how we calculate FOT tax, but it is now fixed in v2-sdk and this should make this code cleaner

### DIFF
--- a/src/providers/v2/quote-provider.ts
+++ b/src/providers/v2/quote-provider.ts
@@ -1,14 +1,10 @@
 import { BigNumber } from '@ethersproject/bignumber';
 import { TradeType } from '@uniswap/sdk-core';
-import {
-  InsufficientInputAmountError,
-  InsufficientReservesError,
-} from '@uniswap/v2-sdk';
+import { InsufficientInputAmountError, InsufficientReservesError } from '@uniswap/v2-sdk';
 
 import { V2Route } from '../../routers/router';
 import { CurrencyAmount } from '../../util/amounts';
 import { log } from '../../util/log';
-import { metric, MetricLoggerUnit } from '../../util/metric';
 import { routeToString } from '../../util/routes';
 import { ProviderConfig } from '../provider';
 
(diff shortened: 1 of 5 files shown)


src/providers/v2/quote-provider.ts
```diff
@@ -157,54 +103,11 @@ export class V2QuoteProvider implements IV2QuoteProvider {
...
-                    inputAmount.quotient
-                  );
-                  [inputAmount] = pair.getInputAmount(inputAmountWithBuyFeeBps);
-                } else {
-                  [inputAmount] = pair.getInputAmount(inputAmount);
-                }
-              } else {
-                [inputAmount] = pair.getInputAmount(inputAmount);
-              }
+            for (const pair of route.pairs.reverse()) {
```
mikeki, line 106: Please help me double check that this for loop remains the same

src/providers/v2/quote-provider.ts
```diff
@@ -92,62 +90,10 @@ export class V2QuoteProvider implements IV2QuoteProvider {
...
-                  const [outputAmountNew] = pair.getOutputAmount(outputAmount);
-                  outputAmount = outputAmountNew;
-                }
-              } else {
-                const [outputAmountNew] = pair.getOutputAmount(outputAmount);
-                outputAmount = outputAmountNew;
-              }
+              [outputAmount] = pair.getOutputAmount(
+                outputAmount,
+                providerConfig.enableFeeOnTransferFeeFetching ?? false
```
jsy1218, line 95: nit - this reads a little weird. `providerConfig.enableFeeOnTransferFeeFetching === true` seems more direct.

src/providers/v2/quote-provider.ts
jsy1218, line 106: why is there a reverse?

jsy1218: aw I see at this point, prefer to retain the same code:

```
for (let i = route.pairs.length - 1; i >= 0; i--) {
              const pair = route.pairs[i]!;
}
```

jsy1218 (approved):

jsy1218 (approved):

src/providers/v2/quote-provider.ts
jsy1218: Looks fine!

PR #423, opened 2023-09-30
Fix FOT first token not having tax info

mikeki:
First token of FOT doesn't contain the tax info, we need to pull it from the pair.

### DIFF
--- a/src/providers/v2/quote-provider.ts
+++ b/src/providers/v2/quote-provider.ts
@@ -87,7 +87,11 @@ export class V2QuoteProvider implements IV2QuoteProvider {
       for (const amount of amounts) {
         try {
           if (tradeType == TradeType.EXACT_INPUT) {
-            let outputAmount = amount.wrapped;
+            const firstPair = route.pairs[0]!;
+            let outputAmount = CurrencyAmount.fromRawAmount(
+              firstPair.token0.equals(amount.wrapped.currency) ? firstPair.token0 : firstPair.token1,
+              amount.wrapped.quotient
+            );
 
             for (const pair of route.pairs) {
               [outputAmount] = pair.getOutputAmount(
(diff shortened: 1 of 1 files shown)


src/providers/v2/quote-provider.ts
```diff
@@ -87,7 +87,11 @@ export class V2QuoteProvider implements IV2QuoteProvider {
       for (const amount of amounts) {
         try {
           if (tradeType == TradeType.EXACT_INPUT) {
-            let outputAmount = amount.wrapped;
+            const firstPair = route.pairs[0]!;
+            let outputAmount = CurrencyAmount.fromRawAmount(
```
jsy1218, line 91: I've been thinking if we can inline this special logic inside v2-sdk? 

```
private derivePercentAfterSellFees(inputAmount: CurrencyAmount<Token>): Percent {
    const sellFeeBips = firstPair.token0.equals(amount.wrapped.currency) ? firstPair.token0.wrapped.sellFeeBps : firstPair.token1.wrapped.sellFeeBps;
    if (sellFeeBips?.gt(BigNumber.from(0))) {
      return ONE_HUNDRED_PERCENT.subtract(new Percent(JSBI.BigInt(sellFeeBips)).divide(BASIS_POINTS))
    } else {
      return ZERO_PERCENT
    }
  }
```

seems cleaner this way, and we don't have to instantiate CurrencyAmount object one time just for this special case.

jsy1218 (approved):

jsy1218:
Sorry didn't realize Tests / Integration Tests - Quote For Other Networks have failures due to Tenderly simulation, might warrant some investigations.. Withdraw approvals for now..

Also want to spend some time fixing [quote-provider.test.ts](https://github.com/Uniswap/smart-order-router/blob/main/test/unit/providers/v2/quote-provider.test.ts), because this unit test is supposed to catch this bug at the earliest feedback cycle.

jsy1218:
Let's close this in favor of https://github.com/Uniswap/smart-order-router/pull/432

PR #425, opened 2023-10-09
feat: support portion

ghost:
feature

After length discussions, we determined majority of https://github.com/Uniswap/unified-routing-api/pull/269 should be in [smart-order-router](https://github.com/Uniswap/smart-order-router). [unified-routing-api](https://github.com/Uniswap/unified-routing-api) should mimimize to call portion and pass down bips and recipient to routing-api. 

[smart-order-router](https://github.com/Uniswap/smart-order-router) migrated majority of the implementations from https://github.com/Uniswap/unified-routing-api/pull/269

- **Other information**:
- [x] https://github.com/Uniswap/smart-order-router/compare/support_portion?expand=1#diff-2fe4a092f36b8091262f7035638dc00530816382f3a1cca86927a6fd712e9c04R1284-R1289 is still an open question.
- [x] add unit test and integ-test

### DIFF
--- a/src/routers/router.ts
+++ b/src/routers/router.ts
@@ -49,7 +49,8 @@ export type SwapRoute = {
    * this estimate to be in terms of the quote token, and subtracting that from the quote.
    * i.e. quoteGasAdjusted = quote - estimatedGasUsedQuoteToken
    */
-  quoteGasAdjusted: CurrencyAmount;  /**
+  quoteGasAdjusted: CurrencyAmount;
+  /**
    * The quote adjusted for the estimated gas used by the swap as well as the portion amount, if applicable.
    * This is computed by estimating the amount of gas used by the swap, converting
    * this estimate to be in terms of the quote token, and subtracting that from the quote.


src/routers/alpha-router/alpha-router.ts
```diff
@@ -948,6 +963,17 @@ export class AlphaRouter
     swapConfig?: SwapOptions,
     partialRoutingConfig: Partial<AlphaRouterConfig> = {}
   ): Promise<SwapRoute | null> {
+    if (tradeType === TradeType.EXACT_OUTPUT) {
```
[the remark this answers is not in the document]
cgkol: or i guess its fine because we return the 0 amount?

cgkol (commented):
We definitely need test for this; LG overall though!

src/providers/portion-provider.ts
```diff
@@ -79,61 +98,57 @@ export class PortionProvider implements IPortionProvider {
...
-        return quoteGasAdjusted.subtract(portionAmount);
+        return quote;
       case TradeType.EXACT_OUTPUT:
-        return this.getQuoteGasAndPortionAdjustedForExactOut(
-          portionAmount,
-          quote,
-          quoteGasAdjusted,
-          amount
-        );
+        return quote.subtract(portionQuoteAmount);
```
cgkol, line 177: thats because we went with the alternative approach here right? https://www.notion.so/uniswaplabs/Fees-Fields-Handshake-27f52f89b0424a7c88bb834d1757af40?d=4cce125efd73410f81ab001520df49f3#4b16f468af2347f5b1411ab49b8a5d91

src/providers/portion-provider.ts
```diff
@@ -79,61 +98,57 @@ export class PortionProvider implements IPortionProvider {
...
+    tradeType: TradeType,
+    quoteGasAdjusted: CurrencyAmount,
+    portionQuoteAmount: CurrencyAmount
   ): CurrencyAmount {
-    // we have a simple heuristic to estimate the quote gas and portion adjusted for exact out
+    switch (tradeType) {
+      case TradeType.EXACT_INPUT:
+        return quoteGasAdjusted;
+      case TradeType.EXACT_OUTPUT:
+        return quoteGasAdjusted.subtract(portionQuoteAmount);
```
cgkol, line 192: i would add a comment explaining this honeslty; but other than that looks good

src/providers/portion-provider.ts
```diff
@@ -79,61 +98,57 @@ export class PortionProvider implements IPortionProvider {
...
+  getQuoteGasAndPortionAdjusted(
+    tradeType: TradeType,
+    quoteGasAdjusted: CurrencyAmount,
+    portionAmount: CurrencyAmount
+  ): CurrencyAmount {
+    switch (tradeType) {
+      case TradeType.EXACT_INPUT:
+        return quoteGasAdjusted.subtract(portionAmount);
+      case TradeType.EXACT_OUTPUT:
+        return quoteGasAdjusted;
```
cgkol, line 207: so now this `quoteGasAdjusted` is actually `quoteGasAndPortionAdjusted` right? this is also kind of a misnomer at this point? Anything we can do about this?

cgkol: maybe call the function already with `quoteGasAdjusted` - `portionQuoteAmount` or somethinig?

cgkol: litearlly on the definition you say 

```
For exact out, quote gas adjusted amount is the best swap quote gas adjusted amount minus the portion quote token amount.
```

but actually you dont pass in that right now

src/routers/alpha-router/alpha-router.ts
```diff
@@ -948,6 +963,24 @@ export class AlphaRouter
...
+        swapConfig
+      );
+      if (portionAmount && portionAmount.greaterThan(ZERO)) {
+        // In case of exact out swap, before we route, we need to make sure that the
+        // token out amount accounts for flat portion, and token in amount after the best swap route contains the token in equivalent of portion.
+        // In other words, in case a pool's LP fee bps is lower than the portion bps (0.01%/0.05% for v3), a pool can go insolvency.
+        // This is because instead of the swapper being responsible for the portion,
+        // the pool instead gets responsible for the portion.
+        // The addition below avoids that situation.
+        amount = amount.add(portionAmount);
```
mikeki, line 980: so this is the exactOut amount after adding the portion, right? because UniversalRouter will then take the portion after the swap. is that right?

src/routers/router.ts
```diff
@@ -49,7 +49,14 @@ export type SwapRoute = {
    * this estimate to be in terms of the quote token, and subtracting that from the quote.
    * i.e. quoteGasAdjusted = quote - estimatedGasUsedQuoteToken
    */
-  quoteGasAdjusted: CurrencyAmount;
+  quoteGasAdjusted: CurrencyAmount;  /**
```
mikeki, line 52: nit: comment is in the same line

mikeki (approved):

PR #429, opened 2023-10-12
fix: route output amount needs to subtract the portion amount only for exact in

ghost:
Bug fix

Right now SOR does not have route output amount subtract the portion amount. The result is token out amount is the same as before as constructed as calldata in the universal router commands.

portion provider will in-place update the routeWithQuotes object to update the quote field with the portion amount. This is only applicable to exact in.

- **Other information**:

URA integ-test can pass on my local: https://app.warp.dev/block/SwN3m90w2VGdLmM93lqGHD

### DIFF
--- a/test/unit/providers/portion-provider.test.ts
+++ b/test/unit/providers/portion-provider.test.ts
@@ -1,17 +1,30 @@
-import { Currency, CurrencyAmount, Fraction, Percent, Token, TradeType } from '@uniswap/sdk-core';
-import { parseAmount, SwapOptions, SwapType } from '../../../src';
+import { BigNumber } from '@ethersproject/bignumber';
+import {
+  Currency,
+  CurrencyAmount,
+  Fraction,
+  Percent,
+  Token,
+  TradeType,
+} from '@uniswap/sdk-core';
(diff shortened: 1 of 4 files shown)


mikeki (approved):

jsy1218 (approved):

test/unit/providers/portion-provider.test.ts
```diff
@@ -138,4 +151,99 @@ describe('portion provider', () => {
...
+          recipient: FLAT_PORTION.recipient
+        }
+      }
+      const oneHundredPercent = new Percent(1);
+
+      const routesWithQuotePortionAdjusted = portionProvider.getRouteWithQuotePortionAdjusted(TradeType.EXACT_INPUT, routesWithValidQuotes, swapParams);
+
+      routesWithQuotePortionAdjusted.forEach((routeWithQuotePortionAdjusted) => {
+        if (routeWithQuotePortionAdjusted instanceof V2RouteWithValidQuote) {
+          expect(routeWithQuotePortionAdjusted.quote.quotient.toString()).toEqual(oneHundredPercent.subtract(new Percent(FLAT_PORTION.bips, 10_000)).multiply(20).quotient.toString())
```
cgkol, line 190: nit: we could maybe even hardcode the 20, 50, 60 per test

cgkol (approved):
amazing!

PR #438, opened 2023-11-14
Fix: improve tenderly connection

jsy1218:
Bug fix

Currently we don't have monitoring against the downstream tenderly http call success rate

We are logging the http status from tenderly against the tenderly request count.

- **Other information**:
This was a follow-up item from the [incident](https://www.notion.so/uniswaplabs/New-tokens-swap-outages-on-Wallet-due-to-Tenderly-enforcing-HTTPS-aad4719087f0407d8a9b21f0137fd070?pvs=4#86b9072133b949a6b9e636dfcca1e76e). We cannot alert against tenderly simulation failures, because those could be genuine failures due to user wallet, market conditions, etc. However non-2xx HTTP status code returned from tenderly should be a good success rate metric.

Tested in local routing-api, and can see new metrics:
Screenshot 2023-11-14 at 3 43 37 PM

### DIFF
--- a/src/providers/tenderly-simulation-provider.ts
+++ b/src/providers/tenderly-simulation-provider.ts
@@ -35,6 +35,8 @@ import {
 import { IV2PoolProvider } from './v2/pool-provider';
 import { ArbitrumGasData, OptimismGasData } from './v3/gas-data-provider';
 import { IV3PoolProvider } from './v3/pool-provider';
+import http from 'http';
+import https from 'https';
 
 export type TenderlyResponseUniversalRouter = {
   config: {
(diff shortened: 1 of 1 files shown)


cgkol (approved):
✅

src/providers/tenderly-simulation-provider.ts
```diff
@@ -396,11 +399,15 @@ export class TenderlySimulator extends Simulator {
...
+
       const before = Date.now()
 
-      const resp = (
+      const { data: resp, status: httpStatus } = (
         await axios.post<TenderlyResponseSwapRouter02>(url, body, opts)
-      ).data;
+      );
+
+      metric.putMetric('TenderlySimulationSwapRouter02ResponseStatus', httpStatus);
```
mikeki, line 410: I think we need to separate these, add the status in the metric name rather than the value

mikeki (approved):

cah4a:
That PR breaks react-native.

```
Bundling failed 11682ms
The package at "node_modules/@uniswap/smart-order-router/build/module/providers/tenderly-simulation-provider.js" attempted to import the Node standard library module "http".
It failed because the native React runtime does not include the Node standard library.
Learn more: https://docs.expo.dev/workflow/using-libraries/#using-third-party-libraries
```

PR #440, opened 2023-11-15
fix: always use tenderly simulation for ETH input

ConjunctiveNormalForm:
- bug fix for UniswapX ETH input flow

  - use eth_estimateGas if input is ETH
  - use actual user balance

  - always use eth_estimateGas for ETH input
  - use Beacon Chain Deposit contract as from address so it always hold enough balance

- **Other information**:


src/providers/tenderly-simulation-provider.ts
```diff
@@ -125,7 +125,7 @@ export class FallbackTenderlySimulator extends Simulator {
       ))
     ) {
       log.info(
-        'Simulating with eth_estimateGas since token is native or approved.'
+        'Simulating with eth_estimateGas since token is approved.'
```
marktoda, line 128: ooh wait can't we always do eth_estimateGas with the deposit address as from? i.e. no need for tenderly in this case

marktoda: like if you keep the conditional above and move the fromAddress update here this should work and save us tenderly costs +  latency

ConjunctiveNormalForm: ah yeah good idea

src/providers/eth-estimate-gas-provider.ts
```diff
@@ -46,10 +46,13 @@ export class EthEstimateGasSimulator extends Simulator {
     providerConfig?: ProviderConfig
   ): Promise<SwapRoute> {
     const currencyIn = route.trade.inputAmount.currency;
+    if (currencyIn.isNative) {
+      fromAddress = BEACON_CHAIN_DEPOSIT_ADDRESS;
```
marktoda, line 50: lil comment here explaining why we do this would be useful

marktoda (approved):

marktoda (approved):

marktoda (approved):

src/providers/tenderly-simulation-provider.ts
```diff
@@ -1,3 +1,6 @@
+import http from 'http';
+import https from 'https';
```
cah4a, line 2: That breaks react-native:

```
Bundling failed 37152ms
The package at "node_modules/@uniswap/smart-order-router/build/main/providers/tenderly-simulation-provider.js" attempted to import the Node standard library module "http".
It failed because the native React runtime does not include the Node standard library.
```

PR #445, opened 2023-11-16
Add support for optional user specified gasToken

zhongeric:
Adding a new feature to gas estimation logic and a general refactor of the gas models to make the code simpler and cleaner.

This PR allows for an optional parameter to be specified in the AlphaRouter config called `gasToken`. The type is initially a string in `AlphaRouterConfig` and will be resolved into a `Token` class before being passed into the gas models. If gasToken is present in the config, during gas estimation we fetch a new native/GasToken pool in parallel to the other pools (native/QuoteToken, native/AmountToken). If the pool fetch was successful, we estimate the gas cost in terms of the `gasToken` and return it as a new, conditionally present response field from the gas models.

This PR also greatly increases the unit test coverage of the gas model + estimation code paths

For testing:
- fetch a USDC -> UNI quote, returning gas in USDC
```
./bin/cli quote --tokenIn 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 --tokenOut 0x1f9840a85d5af5bf1d1762f925bdaddc4201f984 --amount 1000 --exactIn --recipient 0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B --protocols v2,v3 --gasToken 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
```


src/routers/alpha-router/gas-models/gas-model.ts
```diff
@@ -170,13 +183,39 @@ export abstract class IOnChainGasModelFactory {
...
+          ? nativeTokenPool.token0Price
+          : nativeTokenPool.token1Price;
+    let gasCostInTermsOfToken1: CurrencyAmount | undefined;
+    try {
+      // native token is base currency
+      gasCostInTermsOfToken1 = nativeTokenPrice.quote(
+        totalGasCostNativeCurrency
+      ) as CurrencyAmount;
+    } catch (err) {
+      throw err;
```
zhongeric, line 217: TODO: add logging here

jsy1218:
Looks like there's some integ-test failures. Can you rerun and ensure they all pass first?

jsy1218:
What's the purpose of the gas token? Is it to improve the gas estimate logic for UniswapX quote comparison?

src/providers/caching-gas-provider.ts
```diff
@@ -12,7 +12,8 @@ import { GasPrice, IGasPriceProvider } from './gas-price-provider';
  * @class CachingV3SubgraphProvider
```
jsy1218, line 12: All those gas-providers related source files should have no changes in this PR. It feels like the PR isn't fully merged from main.

src/providers/provider.ts
```diff
@@ -1,14 +1,8 @@
...
-
 export type ProviderConfig = {
   /**
    * The block number to use when getting data on-chain.
    */
   blockNumber?: number | Promise<number>;
-  /*
-   * Any additional overhead to add to the gas estimate
-   */
-  additionalGasOverhead?: BigNumber;
```
jsy1218, line 11: Why removing `additionalGasOverhead` though?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1995,17 +2019,33 @@ export class AlphaRouter
...
       : Promise.resolve(null);
 
-    const [usdPool, nativeQuoteTokenV3Pool, nativeAmountTokenV3Pool] =
-      await Promise.all([
-        usdPoolPromise,
-        nativeQuoteTokenV3PoolPromise,
-        nativeAmountTokenV3PoolPromise,
-      ]);
+    const nativeGasTokenV3PoolPromise =
+      providerConfig?.gasToken &&
```
jsy1218, line 2023: nit: I think it can be shorten to `!nativeCurrency .equals(providerConfig?.gasToken)`

src/routers/alpha-router/gas-models/gas-model.ts
```diff
@@ -170,13 +187,31 @@ export abstract class IOnChainGasModelFactory {
...
+// Determines if native currency is token0
+// Gets the native price of the pool, dependent on 0 or 1
+// quotes across the pool
+export const getQuoteThroughNativePool = (
+  chainId: ChainId,
+  nativeTokenAmount: CurrencyAmountRaw<Token>,
+  nativeTokenPool: Pool | Pair
+): CurrencyAmount => {
+  const nativeCurrency = WRAPPED_NATIVE_CURRENCY[chainId];
+  const isToken0 = nativeTokenPool.token0.address == nativeCurrency.address;
```
jsy1218, line 210: lowercase?

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
```diff
@@ -61,55 +62,16 @@ export class MixedRouteHeuristicGasModelFactory extends IOnChainGasModelFactory
...
-        estimateGasCost,
-      };
-    }
-
-    // If the quote token is not in the native currency, we convert the gas cost to be in terms of the quote token.
-    // We do this by getting the highest liquidity <quoteToken>/<nativeCurrency> pool. eg. <quoteToken>/ETH pool.
-    const nativeV3Pool: Pool | null = pools.nativeQuoteTokenV3Pool;
+    const usdPool: Pool = pools.usdPool;
+    const usdToken =
+      usdPool.token0.address == nativeCurrency.address
```
jsy1218, line 68: lowercase? or maybe wrap the nativeCurrency to invoke the `.equals()`

src/util/gas-factory-helpers.ts
jsy1218, line 1: I think the `calculateGasUsed` method in this file also needs to support returning `gasCostInGasToken`. This is used by eth/tenderly provider to provide gas cost in gas tokens in case a simulation is requested.

test/integ/routers/alpha-router/alpha-router.integration.test.ts
jsy1218, line 1: Accompanied with https://github.com/Uniswap/smart-order-router/pull/445/files#r1408462126, I think the integ-test is missing a test case of swap with simulation. If my understanding is correct, in case the tenderly simulation is triggered, `gasCostInGasToken` will not get returned.

zhongeric:
> What's the purpose of the gas token? Is it to improve the gas estimate logic for UniswapX quote comparison?

So this is for the upcoming relayer project, which allows users to pay for gas using ERC20 tokens that are not ETH

src/routers/alpha-router/alpha-router.ts
zhongeric: `.equals` doesn't seem to allow for undefined unfortunately so I think I still have to check for definedness

src/routers/alpha-router/gas-models/gas-model.ts
zhongeric: opted for `Currency.equals`, nice catch!

src/util/gas-factory-helpers.ts
zhongeric: Ah yes good catch. Going to refactor that code path and implement as well!

src/providers/caching-gas-provider.ts
zhongeric: Seems like prettier / eslint changes

test/integ/routers/alpha-router/alpha-router.integration.test.ts
zhongeric: Great catch, missed that code path - fixed in latest version

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1995,17 +2019,33 @@ export class AlphaRouter
...
+      nativeQuoteTokenV3PoolPromise,
+      nativeAmountTokenV3PoolPromise,
+      nativeGasTokenV3PoolPromise,
+    ]);
 
     const pools: LiquidityCalculationPools = {
       usdPool: usdPool,
       nativeQuoteTokenV3Pool: nativeQuoteTokenV3Pool,
       nativeAmountTokenV3Pool: nativeAmountTokenV3Pool,
+      nativeGasTokenV3Pool: nativeGasTokenV3Pool,
```
jsy1218, line 2048: nit: I think the readability will benefit from adding some comments, explaining this new pool, although it's called native, is not to be confused with the native ETH as gas token.

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
```diff
@@ -137,6 +94,44 @@ export class MixedRouteHeuristicGasModelFactory extends IOnChainGasModelFactory
...
+          nativeGasTokenPool
+        );
+      }
+      // if the gasToken is the native currency, we can just use the totalGasCostNativeCurrency
+      else if (providerConfig?.gasToken?.equals(nativeCurrency)) {
+        gasCostInTermsOfGasToken = totalGasCostNativeCurrency;
+      }
+
+      /** ------ MARK: return early if quoteToken is wrapped native currency ------- */
+      if (quoteToken.equals(nativeCurrency)) {
```
jsy1218, line 120: nit: it seems like if `quoteToken.equals(nativeCurrency)` is true, then `providerConfig?.gasToken?.equals(nativeCurrency)` is also true, meaning we can even return earlier.

But I do think this way the readability degrades a bit.

src/util/gas-factory-helpers.ts
```diff
@@ -362,11 +370,12 @@ export function initSwapRouteFromExisting(
   v2PoolProvider: IV2PoolProvider,
   v3PoolProvider: IV3PoolProvider,
   portionProvider: IPortionProvider,
-  quoteGasAdjusted: CurrencyAmount<Currency>,
+  quoteGasAdjusted: CurrencyAmount,
```
jsy1218, line 373: nit: keeping generic does have benefit I think, and `Currency` is a super type of `NativeCurrency` and `Token` right?

jsy1218 (approved):

src/routers/alpha-router/alpha-router.ts
zhongeric: Yep agree, actually changing the naming throughout to be `nativeAnd<token>` to hopefully make it more explicit that it refers to a single V3 pool between the native currency and some other token

src/routers/alpha-router/gas-models/mixedRoute/mixed-route-heuristic-gas-model.ts
zhongeric: mmm no so the gasToken is not always the quoteToken, in fact the main use case of this is to allow for us to quote a route's estimated gas in a third token that is not the tokenIn or tokenOut

src/util/gas-factory-helpers.ts
zhongeric: Seems like our local `CurrencyAmount` is just this:
```
export class CurrencyAmount extends CurrencyAmountRaw<Currency> {}
```
And I initially tried keeping the generic CurrencyAmount here but was getting weird type issues when dealing with wrapped native currency

zhongeric:
Merging!

PR #448, opened 2023-11-18
Chore: add tenderly timeout metrics

jsy1218:
Metrics

We don't track the number of tenderly request timeouts

We start tracking the number of tenderly request timeouts

- **Other information**:
In routing-api prod account, I see a single instance of 3s timeout, and the error message contains the timeout:
Screenshot 2023-11-17 at 4 03 44 PM

From this [guide](https://rapidapi.com/guides/axios-timeouts), it suggests `error.code === 'ECONNABORTED' && error.message.includes('timeout')`. The `Error` library in sor has the following constructs:

```
interface Error {
    name: string;
    message: string;
    stack?: string;
}
```

So I'm opting to check against `error.message` only for timeout

- [x] Reduce timeout to very small number in routing-api, and intentionally trigger the timeout from local routing-api integ-test to test this sor metric change. Verified on my local routing-api AWS account to see the new `TenderlyTimeout` metric, with the corresponding error logging:
Screenshot 2023-11-20 at 10 35 08 AM
Screenshot 2023-11-20 at 10 37 10 AM

### DIFF
--- a/src/providers/tenderly-simulation-provider.ts
+++ b/src/providers/tenderly-simulation-provider.ts
@@ -154,7 +154,7 @@ export class FallbackTenderlySimulator extends Simulator {
         providerConfig
       );
     } catch (err) {
-      log.info({ err: err }, 'Failed to simulate via Tenderly');
+      log.error({ err: err }, 'Failed to simulate via Tenderly');
 
       if (err instanceof Error && err.message.includes('timeout')) {
         metric.putMetric('TenderlySimulationTimeouts', 1, MetricLoggerUnit.Count);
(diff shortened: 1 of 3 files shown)


cgkol (approved):

src/providers/tenderly-simulation-provider.ts
```diff
@@ -155,6 +155,10 @@ export class FallbackTenderlySimulator extends Simulator {
       );
     } catch (err) {
       log.info({ err: err }, 'Failed to simulate via Tenderly');
```
mikeki, line 157: Can we change this to `log.error`?

jsy1218: Changed to `log.error`, and can see the logging with higher level:
Screenshot 2023-11-20 at 10 41 35 AM

PR #459, opened 2023-12-18
fix: eth_feeHistory by passing 'latest' block number, meanwhile preserving the researcher's ability to quote against past number including gas fees

jsy1218:
Bug fix

We have some quote 5xx error, due to `eth_feeHistory failing on mainnet: request beyond head block: requested x_block_number, head y_block_number`. This is due to Infura node pool eventual consistency - not all the node in the same Infura node pool has the same block header.

We pass in 'latest' in case quote request does not pass in the block number. This is to replace the numerical latest block number we pass to infura during `eth_feeHistory` RPC calls.

- **Other information**:

### DIFF
--- a/src/providers/caching-gas-provider.ts
+++ b/src/providers/caching-gas-provider.ts
@@ -43,7 +43,6 @@ export class CachingGasStationProvider extends IGasPriceProvider {
       return cachedGasPrice;
     }
 
-    log.info('Gas station price local cache miss.');
     const gasPrice = await this.gasPriceProvider.getGasPrice(latestBlockNumber, requestBlockNumber);
     await this.cache.set(this.GAS_KEY(this.chainId, targetBlockNumber), gasPrice);
 
--- a/src/providers/caching-gas-provider.ts
+++ b/src/providers/caching-gas-provider.ts
@@ -28,8 +28,11 @@ export class CachingGasStationProvider extends IGasPriceProvider {
     super();
   }
 
-  public override async getGasPrice(requestBlockNumber: number): Promise<GasPrice> {
-    const cachedGasPrice = await this.cache.get(this.GAS_KEY(this.chainId, requestBlockNumber));
+  public override async getGasPrice(latestBlockNumber: number, requestBlockNumber?: number): Promise<GasPrice> {
+    // If block number is specified in the request, we have to use that block number find any potential cache hits.
+    // Otherwise, we can use the latest block number.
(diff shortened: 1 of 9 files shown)


src/providers/caching-gas-provider.ts
```diff
@@ -41,8 +44,8 @@ export class CachingGasStationProvider extends IGasPriceProvider {
     }
 
     log.info('Gas station price local cache miss.');
```
cgkol, line 46: i think we should delete this too btw, we dont need an info here

src/providers/eth-gas-station-info-gas-price-provider.ts
```diff
@@ -29,7 +29,7 @@ export class ETHGasStationInfoProvider extends IGasPriceProvider {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  public override async getGasPrice(_requestBlockNumber: number): Promise<GasPrice> {
+  public override async getGasPrice(_latestBlockNumber: number, _requestBlockNumber?: number): Promise<GasPrice> {
     log.info(`About to get gas prices from gas station ${this.url}`);
```
cgkol, line 33: ame here

src/providers/legacy-gas-price-provider.ts
```diff
@@ -10,7 +10,7 @@ export class LegacyGasPriceProvider extends IGasPriceProvider {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  public override async getGasPrice(_requestBlockNumber: number): Promise<GasPrice> {
+  public override async getGasPrice(_latestBlockNumber: number, _requestBlockNumber?: number): Promise<GasPrice> {
     const gasPriceWei = await this.provider.getGasPrice();
```
cgkol, line 14: woa everywhere

cgkol (approved):
just info comments

src/providers/caching-gas-provider.ts
jsy1218: Agree.. Kinda silly imho.. If we want gas price cache hit/miss ratio, we should log metrics.

mikeki (approved):

src/providers/legacy-gas-price-provider.ts
jsy1218: removed those unnecessary info logging

PR #465, opened 2024-01-03
feat: enable v2 routing on all deployed L2

just-toby:

enables v2 routes on all deployed L2

currenty SOR assumes v2 isn't deployed on all L2

support v2 routes on all deployed L2

- **Other information**:


jsy1218:
Looks like the merge conflict needs to be resolved before Github CI runs the tests.

cgkol (commented):
Shall we run some tests to make sure this works on optmism before we push it?

jsy1218:
> Shall we run some tests to make sure this works on optmism before we push it?

I think we want to conduct some manual testing (and maybe add into integ-test as well):

1. getting the new UR deployed address https://uniswapteam.slack.com/archives/C05MF1J6GR5/p1704913678240789
2. deploy V2 tokens on optimism ourselves (preferably FOT)
3. create FOT/WETH pool
4. add liquidity into FOT/WETH pool
5. run router to make sure it can route through

We are blocked on 1 right now.

jsy1218:
Also I just noticed that somehow for v2 pool-provider, SOR doesn't use [FACTORY_ADDRESS](https://github.com/Uniswap/v2-sdk/blob/main/src/constants.ts#L7) from v2-sdk, but rather computes the pool address directly https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v2/pool-provider.ts#L261. This was probably fine previously, because UR was not deployed for v2 pools. But now with UR with different factory addresses across chains, we might need to change the v2 pool-provider logic.

mikeki (approved):
Overall LGTM, assuming it works

jsy1218:
I noticed v2-sdk [FACTORY_ADDRESS_MAP](https://github.com/Uniswap/v2-sdk/pull/151/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R4) is missing CELO. This explains why the CELO integ-tests are consistently failing. We will need to bump v2-sdk to include re-deployed CELO address.

jsy1218:
> Shall we run some tests to make sure this works on optmism before we push it?

Sadly we aren't able to test on any deployed L2, including Optimism. The reason is because no has added liquidity on V2 deployed L2 yet. For example, the latest [UniswapV2Router02](https://optimistic.etherscan.io/address/0x920b806e40a00e02e7d2b94ffc89860fdaed3640) deployed Optimism shows no one ever invoked addLiquidity/addLiquidityETH of the [write contract](https://optimistic.etherscan.io/address/0x920b806e40a00e02e7d2b94ffc89860fdaed3640#writeContract).

I think we want to merge this first, meanwhile getting to how to add liquidity to test routing.

PR #469, opened 2024-01-09
Optimize get candidate pools

cuongquangnam:

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/functions/get-candidate-pools.ts
+++ b/src/routers/alpha-router/functions/get-candidate-pools.ts
@@ -445,12 +445,12 @@ export async function getV3CandidatePools({
         ? subgraphPool.token1.id
         : subgraphPool.token0.id;
     })
+    .filter((secondHopId: string) => !tokensToAvoidOnSecondHops?.includes(secondHopId.toLowerCase()))
     .flatMap((secondHopId: string) => {
       return _(subgraphPoolsSorted)
         .filter((subgraphPool) => {
           return (
             !poolAddressesSoFar.has(subgraphPool.id) &&
-            !tokensToAvoidOnSecondHops?.includes(secondHopId.toLowerCase()) &&
             (subgraphPool.token0.id == secondHopId ||
               subgraphPool.token1.id == secondHopId)
           );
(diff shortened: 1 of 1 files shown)


src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -445,12 +445,12 @@ export async function getV3CandidatePools({
         ? subgraphPool.token1.id
         : subgraphPool.token0.id;
     })
+    .filter((secondHopId: string) => !tokensToAvoidOnSecondHops?.includes(secondHopId.toLowerCase()))
```
jsy1218, line 448: In theory, this filter logic should make flatMap instantiate less subgraphPool objects at runtime, and thus more performant. In practice, I think this change actually might bring perf overhead, due to the pools characteristics as well as lodash library.

So to give some context, back then, we noticed that lodash instantiates objects per function like `filter` and `flatMap` based on the perf analysis. Because of that, we initial tried to refactor both V2 and V3 subgraph pool selections without lodash https://github.com/Uniswap/smart-order-router/pull/331. For some reason, we saw perf improvement in V2 pool selection, but perf regression in V3 pool selection, hence we reverted back V3 pool selection https://github.com/Uniswap/smart-order-router/pull/340.

Next in V2 pool selection, we optimize the second hops if we have enough candidate pools for routes https://github.com/Uniswap/smart-order-router/pull/371, as well as tracked the loop iterations in metrics https://github.com/Uniswap/smart-order-router/pull/351. It turns out even with all those optimizations, V2 candidate pools were almost all iterated per quote request using our internal AWS metrics:
 
Screenshot 2024-01-10 at 2 47 11 PM

Thus we made the conclusion, that majority of the V2 pools do not have enough liquidities (we use TVL as approximate to TVL in SOR code), so no matter how we optimize to avoid second hops, at runtime, V2 candidate pools will be almost all iterated. Later on when I checked on https://v2.info.uniswap.org/pairs, the liquidities ranking does seem to show most V2 pools don't have high enough liquidities.

For V3, we don't have a metric to track how many tokens avoided the second hops, but from https://info.uniswap.org/home#/pools, I can that majority of the TVLs are concentrated in the top pools as well. This leaves me to think that the after `.filter`, majority of the V3 pools will still be there at runtime, and new pools objects will get instantiated. Then after `flatMap` new pools objects will get instantiated once more.

jsy1218:
We should close the PR given my explanation above.

PR #468, opened 2024-01-06
fix: Arb gas estimate brotli compression calldata byte estimate and estimateGasUsed adding arb L2 gas cost

jsy1218:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fixes

- **What is the current behavior?** (You can also link to an open issue here)
There are 3 bugs:

1. Arbitrum calldata gas cost is based on the original calldata bytes, not brotli compressed bytesize.
2. non-simulated `estimatedGasUsed` does not add the Arbitrum L1 calldata posting fee in terms of the L2 gas price.
3. Arb and OP gas price oracle does not pass the block number to the block tag for historical gas price.

- **What is the new behavior (if this is a feature change)?**
There are 3 fixes:

1. Use Brotli-One compression as an approximate to get the compressed Arbitrum L2->L1 calldata bytesizes.
2. In non-simulated `estimatedGasUsed` calculation, add back the Arbitrum L2->L1 gas cost in terms of L2 Arb gas price.
3. Pass block number to the block tag, so that the historical quote with gas adjusted will be more accurate on Arb and OP network.

- **Other information**:
https://github.com/Uniswap/smart-order-router/pull/466#issuecomment-1878547849 shows the updated `estimatedGasUsed` under two CLIs. With all 3 fixes, I'm able to replicate the same gas estimate under that historical block:
```
$./bin/cli quote --tokenIn 0xfd086bc7cd5c481dcc9c85ebe478a1c0b69fcbb9 --tokenOut 0x82af49447d8a07e3bd95bd0d56f35241523fbab1 --amount 20000 --exactIn --minSplits 1 --router alpha --chainId 42161 --recipient 0xb38e8c17e38363af6ebdcb3dae12e0243582891d --simulate --requestBlockNumber 167333306
Best Route:
[V3] 80.00% = USDT -- 0.05% [0x641C00A822e8b671738d32a431a4Fb6074E5c79d] --> WETH, [V3] 20.00% = USDT -- 0.01% [0x8c9D230D45d6CfeE39a6680Fb7CB7E8DE7Ea8E71] --> USDC -- 0.05% [0xC31E54c7a869B9FcBEcc14363CF510d1c41fa443] --> WETH
	Raw Quote Exact In:
		8.91
	Gas Adjusted Quote In:
		8.91

Gas Used Quote Token: 0.000137
Gas Used USD: 0.308263
[...]

PR #475, opened 2024-01-17
chore: check gas estimate across L2s

jsy1218:
Follow-up from https://github.com/Uniswap/smart-order-router/pull/468/files#diff-b1a09e5a7156f19225eba6ca4e2e78d0942580ff1118bab472bf0ea3b1294904R3581

- **Other information**:


cgkol (approved):
Amazing! gas estimates should never be off again :)

have we tested that it works? lol

jsy1218:
> Amazing! gas estimates should never be off again :)
> 
> have we tested that it works? lol

Yep tested the gas estimate per chain! Obviously we only cover the basic native currency -> erc20 token swap, so the [GAS_ESTIMATE_DEVIATION_PERCENT](https://github.com/Uniswap/smart-order-router/pull/475/files#diff-b1a09e5a7156f19225eba6ca4e2e78d0942580ff1118bab472bf0ea3b1294904R101) per chain is kinda arbitrary experiment. But it does give us a starting benchmark on how off our gas estimate is per chain :)

PR #485, opened 2024-02-07
feat: add native usdc arbitrum, and usdc/weth arbitrum v2 swap integ-test

jsy1218:
feature

We disabled arbitrum v2 support.

We are enabling arbitrum v2 support. Also the USDC/WETH pool on arbitrum uses native USDC, so we are adding that support.

- **Other information**:


src/routers/alpha-router/alpha-router.ts
```diff
@@ -1713,6 +1719,7 @@ export class AlphaRouter
     v3GasModel: IGasModel<V3RouteWithValidQuote>,
     mixedRouteGasModel: IGasModel<MixedRouteWithValidQuote>,
     gasPriceWei: BigNumber,
+    v2GasModel?: IGasModel<V2RouteWithValidQuote>,
```
mikeki, line 1722: is this ever optional?

mikeki: oh I see now, it is

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -17,8 +17,8 @@ import { AlphaRouterConfig } from '../alpha-router';
 import { IGasModel, L1ToL2GasCosts, usdGasTokensByChain } from '../gas-models';
 
 import {
-  RouteWithValidQuote,
-  V3RouteWithValidQuote,
+  RouteWithValidQuote, V2RouteWithValidQuote,
+  V3RouteWithValidQuote
```
mikeki, line 21: lint?

mikeki (approved):
I'm not an expert on the gas calculations, make sure we get a reviewer from somebody else with more experience on that.

jsy1218:
> I'm not an expert on the gas calculations, make sure we get a reviewer from somebody else with more experience on that.

Exactly, @zhongeric agreed to review as well

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2011,7 +2019,7 @@ export class AlphaRouter
     quoteToken: Token,
     providerConfig?: GasModelProviderConfig
   ): Promise<
-    [IGasModel<V3RouteWithValidQuote>, IGasModel<MixedRouteWithValidQuote>]
+    [IGasModel<V2RouteWithValidQuote> | undefined, IGasModel<V3RouteWithValidQuote>, IGasModel<MixedRouteWithValidQuote>]
```
zhongeric, line 2022: Especially now that it's possible that a V2 model isn't available, I'd rather us change this return type to 
```
{
v2GasModel?: IGasModel<V2RouteWithValidQuote>,
v3GasModel?: ...
}
```
Which provides cleaner restructuring IMO than arrays especially when dealing with optional values

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -451,16 +462,47 @@ export async function getBestSwapRouteBy(
...
+          if (gasCostsL1ToL2.gasCostL1USD.currency.equals(v2GasCostL1.gasCostL1USD.currency)) {
+            gasCostsL1ToL2.gasCostL1USD = gasCostsL1ToL2.gasCostL1USD.add(v2GasCostL1.gasCostL1USD);
+          } else {
+            gasCostsL1ToL2.gasCostL1USD = v2GasCostL1.gasCostL1USD;
+          }
+          if (gasCostsL1ToL2.gasCostL1QuoteToken.currency.equals(v2GasCostL1.gasCostL1QuoteToken.currency)) {
+            gasCostsL1ToL2.gasCostL1QuoteToken = gasCostsL1ToL2.gasCostL1QuoteToken.add(v2GasCostL1.gasCostL1QuoteToken);
+          } else {
+            gasCostsL1ToL2.gasCostL1QuoteToken = v2GasCostL1.gasCostL1QuoteToken;
+          }
```
zhongeric, line 483: Would appreciate some comments on the logic applied here - not sure why we're comparing the currencies for gasCostL1USD.

src/routers/alpha-router/gas-models/v2/v2-heuristic-gas-model.ts
```diff
@@ -90,6 +101,25 @@ export class V2HeuristicGasModelFactory extends IV2GasModelFactory {
...
+      gasCostL1USD: CurrencyAmount;
+      gasCostL1QuoteToken: CurrencyAmount;
+    }> => {
+      return await calculateL1GasFeesHelper(
+        route,
+        chainId,
+        usdPool,
+        token,
+        !token.equals(WRAPPED_NATIVE_CURRENCY[chainId]) ?
+          await getV2NativePool(token, poolProvider, providerConfig) : null,
```
zhongeric, line 118: nit: would prefer this to not be inline

src/util/chains.ts
```diff
@@ -30,11 +30,7 @@ export const V2_SUPPORTED = [
   ChainId.MAINNET,
   ChainId.GOERLI,
   ChainId.SEPOLIA,
-  ChainId.OPTIMISM,
-  ChainId.POLYGON,
-  ChainId.BNB,
-  ChainId.AVALANCHE,
-  ChainId.BASE,
```
zhongeric, line 37: just want to make sure this is intended

src/util/gas-factory-helpers.ts
```diff
@@ -543,3 +544,161 @@ export function initSwapRouteFromExisting(
...
+  let l1Used = BigNumber.from(0);
+  let l1FeeInWei = BigNumber.from(0);
+  let gasUsedL1OnL2 = BigNumber.from(0);
+  const opStackChains = [
+    ChainId.OPTIMISM,
+    ChainId.OPTIMISM_GOERLI,
+    ChainId.OPTIMISM_SEPOLIA,
+    ChainId.BASE,
+    ChainId.BASE_GOERLI,
+  ];
```
zhongeric, line 576: would prefer this in constants file and naming to be something like `L2_FEE_CHAINS` or something more generic

src/util/gas-factory-helpers.ts
```diff
@@ -543,3 +544,161 @@ export function initSwapRouteFromExisting(
...
+  gasCostL1USD: CurrencyAmount;
+  gasCostL1QuoteToken: CurrencyAmount;
+}> => {
+  const swapOptions: SwapOptionsUniversalRouter = {
+    type: SwapType.UNIVERSAL_ROUTER,
+    recipient: '0x0000000000000000000000000000000000000001',
+    deadlineOrPreviousBlockhash: 100,
+    slippageTolerance: new Percent(5, 10_000),
+  };
+  let l1Used = BigNumber.from(0);
```
zhongeric, line 567: nit: naming is a little hard to follow here 😅 , can we do something like `mainnetGasUsed`?

src/util/gas-factory-helpers.ts
```diff
@@ -543,3 +544,161 @@ export function initSwapRouteFromExisting(
...
+      );
+      gasCostL1QuoteToken = CurrencyAmount.fromRawAmount(quoteToken, 0);
+    } else {
+      const nativeTokenPrice =
+        nativePool.token0.address == nativeCurrency.address
+          ? nativePool.token0Price
+          : nativePool.token1Price;
+      gasCostL1QuoteToken = nativeTokenPrice.quote(costNativeCurrency);
+    }
+  }
```
zhongeric, line 628: Can you use the helpers `getQuoteThroughNativePool` that we use elsewhere for this? It'll handle token0/1 and quote across a supplied pool

src/util/gas-factory-helpers.ts
```diff
@@ -543,3 +544,161 @@ export function initSwapRouteFromExisting(
...
+      ChainId.OPTIMISM
+    ).calldata;
+    const l1GasUsed = getL2ToL1GasUsed(data, overhead, chainId);
+    // l1BaseFee is L1 Gas Price on etherscan
+    const l1Fee = l1GasUsed.mul(l1BaseFee);
+    const unscaled = l1Fee.mul(scalar);
+    // scaled = unscaled / (10 ** decimals)
+    const scaledConversion = BigNumber.from(10).pow(decimals);
+    const scaled = unscaled.div(scaledConversion);
+    // TODO: also return the gasUsedL1OnL2 because the final estimateGasUsed should include L1 calldata posting fee
```
zhongeric, line 674: todo?

zhongeric (commented):
looking good! left some comments

src/util/chains.ts
jsy1218: Yep this is intended. @adriantrustman will add liquidity on other L2s. Then in routing-api, we also enable there as well https://github.com/Uniswap/smart-order-router/blob/b970aedfec8a9509f9e22f14cc5c11be54d47b35/src/routers/alpha-router/alpha-router.ts#L832

src/routers/alpha-router/functions/best-swap-route.ts
jsy1218: Yep, because it's possible that gasCostL1QuoteToken and gasCostL1QuoteToken have different currencies

zhongeric: wdym? those are the same variable

src/routers/alpha-router/alpha-router.ts
```diff
@@ -138,11 +138,11 @@ import {
   V3CandidatePools,
 } from './functions/get-candidate-pools';
 import {
-  GasModelProviderConfig,
+  GasModelProviderConfig, GasModelType,
```
zhongeric, line 141: nit: prettier?

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -465,12 +466,18 @@ export async function getBestSwapRouteBy(
...
     } else {
+      // Now after v2 deploy everywhere, it's possible to to have v2 and v3 routes on L2 swap.
+      // So that we need to add both gas cost one by one
       const v2Routes = bestSwap.filter((routes) => routes.protocol === Protocol.V2);
       if (v2Routes.length > 0 && V2_SUPPORTED.includes(chainId)) {
         if (v2GasModel) {
           const v2GasCostL1 = await v2GasModel.calculateL1GasFees!(v2Routes as V2RouteWithValidQuote[]);
           gasCostsL1ToL2.gasUsedL1 = gasCostsL1ToL2.gasUsedL1.add(v2GasCostL1.gasUsedL1);
           gasCostsL1ToL2.gasUsedL1OnL2 = gasCostsL1ToL2.gasUsedL1OnL2.add(v2GasCostL1.gasUsedL1OnL2);
+          // It's possible that the gasCostL1USD currency is different from gasCostL1USD
```
zhongeric, line 477: Sorry I still don't follow what this means here. Is this for when the usdGasToken on mainnet is different from the usdGasToken on the L2? And so in that case we just use the calculated L1 gas fees? The comment is specifying the same variable names so it's hard to follow

src/routers/alpha-router/functions/best-swap-route.ts
```diff
@@ -465,12 +466,18 @@ export async function getBestSwapRouteBy(
...
       throw new Error("Can't compute L1 gas fees.");
     } else {
+      // Now after v2 deploy everywhere, it's possible to to have v2 and v3 routes on L2 swap.
+      // So that we need to add both gas cost one by one
       const v2Routes = bestSwap.filter((routes) => routes.protocol === Protocol.V2);
       if (v2Routes.length > 0 && V2_SUPPORTED.includes(chainId)) {
         if (v2GasModel) {
           const v2GasCostL1 = await v2GasModel.calculateL1GasFees!(v2Routes as V2RouteWithValidQuote[]);
           gasCostsL1ToL2.gasUsedL1 = gasCostsL1ToL2.gasUsedL1.add(v2GasCostL1.gasUsedL1);
           gasCostsL1ToL2.gasUsedL1OnL2 = gasCostsL1ToL2.gasUsedL1OnL2.add(v2GasCostL1.gasUsedL1OnL2);
```
zhongeric, line 476: This naming is very hard to follow and I know that some of it was already like this but would rly appreciate variable names that are more intuitive. Maybe destructing the return value from `v2GasModel.calculateL1GasFees` and instead of doing `.add` you can probably just set `gasCostsL1ToL2.gasUsedL1  = gasUsedL1`?

zhongeric: it seems like there is a lot more logic happening here than there actually is, making it harder to review

src/routers/alpha-router/functions/best-swap-route.ts
jsy1218: I added more detailed explanations https://github.com/Uniswap/smart-order-router/pull/485/files#diff-bf0f39dc98ca22700a5c33efc44d6acf65a840bcface7d39297ffa99d20e9478R507-R521

src/routers/alpha-router/functions/best-swap-route.ts
jsy1218: added more explanations https://github.com/Uniswap/smart-order-router/pull/485/files#diff-bf0f39dc98ca22700a5c33efc44d6acf65a840bcface7d39297ffa99d20e9478R479-R483

tbh, meanwhile I can see the path to full readability (by refactoring v2 and v3 gas models for the L2toL1 security fee part), I don't think the time permits for that amount of refactoring.

zhongeric (approved):

PR #481, opened 2024-01-29
Bump uniswap sdk core to 4.09 and add correct base swap router address

sismanis:
The swap router address for base is different for other chains so the function needs to account for that. Also, @uniswap/sdk-core needs to be bumped to 4.0.9 where the WETH9 contract for Base was added otherwise any pools with WETH on base will not work in this SDK with the swap router.

- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
This will fix bugs with using base with the smart order router. Currently does not work with the current version of the sdk and is fixed with these changes

- **What is the current behavior?** (You can also link to an open issue here)
The current behavior will throw an error when requesting a swap router quote on base

- **What is the new behavior (if this is a feature change)?**
The swap router will work correctly and quotes are able to be fetched using the SDK

- **Other information**:

PR #505, opened 2024-02-23
fix: bump sdk-core and v2-sdk and correct base swap router address

jsy1218:
Bug fix

We previously attempted to fix swap router address on base https://github.com/Uniswap/smart-order-router/pull/481, but the change had regression.

We [fixed](https://github.com/Uniswap/sdk-core/pull/120) the [regression](https://github.com/Uniswap/sdk-core/pull/112) in sdk-core, so that it should work this time.

- **Other information**:
Also releasing
https://github.com/Uniswap/sdk-core/pull/122
https://github.com/Uniswap/v2-sdk/pull/163


jsy1218:
@sismanis FYI shipping this fix again.

PR #507, opened 2024-02-27
add sepolia to v2 supported array

jsy1218:
Bug fix

V2 swap is not supported on Sepolia

We already deployed V2 universal router and factory to Sepolia, we can support Sepolia on L2.

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -2089,7 +2089,7 @@ export class AlphaRouter
           token: quoteToken,
           l2GasDataProvider: this.l2GasDataProvider,
           providerConfig: providerConfig,
-        })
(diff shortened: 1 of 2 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -2089,7 +2089,7 @@ export class AlphaRouter
           token: quoteToken,
           l2GasDataProvider: this.l2GasDataProvider,
           providerConfig: providerConfig,
-        })
+        }).catch(_ => undefined) // If v2 model throws uncaught exception, we return undefined v2 gas model, so there's a chance v3 route can go through
```
mikeki, line 2092: I didn't know underscore was a thing here

mikeki (approved):

src/routers/alpha-router/alpha-router.ts
jsy1218: yea, it's a way to avoid lint error

PR #509, opened 2024-02-29
Bug Fix: Remove Sepolia from v2 support

Rachel-Eichenberger:
Fix to remove Sepolia from v2 support until the new v2 liquidity swapRouter02 launched on Sepolia


jsy1218 (approved):
Let's ship this tmrw during EST business hour

PR #524, opened 2024-04-11
feat: bump sdk to beta

jsy1218:
feature

we are using the published sdks from each sdk repo

we plan to use the sdk monorepo, but the published sdks format will remain the same

- **Other information**:


jsy1218:
This PR is to merely trigger integ-tests and make sure they can pass. Now it can close.

PR #534, opened 2024-04-17
fix: multicall return gas used in case of failure

jsy1218:
feature

Meanwhile we are comparing the current quoter quotes vs new view-only quoter quotes in prod in shadow sampling for mainnet at 1%, we encounter a case, where we have some quotes that are failing due to the out-of-gas from multicall (e.g. this [simulation](https://www.tdly.co/shared/simulation/39de04bb-d794-43b0-aed2-8ea31957ca77)), but from routing-api side, we don't have a good way to programmatically determine this is the case, within [compareQuotes](https://github.com/Uniswap/routing-api/blob/d05f590db3facdc76850fb3ade374b17f485d315/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts#L213). 

This is the case where some quotes from multicalls fail, but overall success rate is still above the [threshold](https://github.com/Uniswap/smart-order-router/blob/19afca6760c5d346ef1d0c20206c3e6758fee0e5/src/providers/on-chain-quote-provider.ts#L496). This is different from the entire multicall failures due to out of gas. In that case, on-chain quote provider will throw [ProviderOutOfGasError](https://github.com/Uniswap/smart-order-router/blob/19afca6760c5d346ef1d0c20206c3e6758fee0e5/src/providers/on-chain-quote-provider.ts#L544), and routing-api can generalize to where [currentQuoterSucceeded](https://github.com/Uniswap/routing-api/blob/d05f590db3facdc76850fb3ade374b17f485d315/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts#L81) = false

Within SOR, we will return both gasUsed and gasLimit per call from the multicalls, so that routing-api can calculate that the difference between gasUsed and gasLimit may be small enough to be due to out of gas.

- **Other information**:
This works in tendom with routing-api changes.

### DIFF
--- a/src/providers/multicall-provider.ts
+++ b/src/providers/multicall-provider.ts
@@ -47,6 +47,7 @@ export type SuccessResult<TReturn> = {
 export type FailResult = {
   success: false;
   returnData: string;
+  gasUsed?: BigNumber;
 };
 
 export type Result<TReturn> = SuccessResult<TReturn> | FailResult;
--- a/src/providers/multicall-uniswap-provider.ts
+++ b/src/providers/multicall-uniswap-provider.ts
@@ -196,6 +196,7 @@ export class UniswapMulticallProvider extends IMulticallProvider<UniswapMultical
         results.push({
           success: false,
           returnData,
+          gasUsed,
         });
         continue;
       }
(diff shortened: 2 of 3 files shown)


cgkol (approved):
When are we planning on doing this? 
```
 so that routing-api can calculate that the difference between gasUsed and gasLimit may be small enough to be due to out of gas
 ```

jsy1218:
> When are we planning on doing this? 
> ```
>  so that routing-api can calculate that the difference between gasUsed and gasLimit may be small enough to be due to out of gas
>  ```
> 
> 

Sometime today. I have the routing-api draft [PR](https://github.com/Uniswap/routing-api/pull/593/files), just need to add some test

PR #540, opened 2024-04-19
chore: token out with fot fix

mikeki:


src/providers/portion-provider.ts
```diff
@@ -202,9 +204,14 @@ export class PortionProvider implements IPortionProvider {
     }
 
     return routeWithQuotes.map((routeWithQuote) => {
+      const tokenOut =
+        routeWithQuote.tokenPath[routeWithQuote.tokenPath.length - 1];
```
jsy1218, line 208: nit: do we want to be defensive and check tokenPath is not empty? Although in theory it should never be empty if it can be here.

src/providers/portion-provider.ts
```diff
@@ -202,9 +204,14 @@ export class PortionProvider implements IPortionProvider {
...
+      const tokenOutHasFot =
+        tokenOut && tokenOut.buyFeeBps && tokenOut.buyFeeBps.gt(0);
```
jsy1218, line 210: nit: `tokenOut.buyFeeBps?.gt(0)`

src/providers/portion-provider.ts
mikeki: I think that if the tokenPath was empty we would have more issues here than just calculating the tokenOut, right? Checking if its non empty would create a fork in here for the empty and non empty path, do you feel strongly about adding that check? I could do it if you do, it just feels like if it were empty it would be caught somewhere else

jsy1218 (approved):

jsy1218 (approved):

PR #547, opened 2024-04-21
fix: improve quote provider logging & update sepolia quoter address

jsy1218:
Bug fix

Our quote provider logging doesn't differentiate between mixed quoter and v3 quoter. We should differentiate them for better perf analysis.
Also the sepolia quoter address has to be updated because it was using the wrong v3 factory address.

Enrich the quote provider logging with the mixed quoter vs v3 quoter.

- **Other information**:

### DIFF
--- a/src/providers/on-chain-quote-provider.ts
+++ b/src/providers/on-chain-quote-provider.ts
@@ -8,7 +8,7 @@ import {
 import { ChainId } from '@uniswap/sdk-core';
 import { encodeRouteToPath } from '@uniswap/v3-sdk';
 import retry, { Options as RetryOptions } from 'async-retry';
-import _, from 'lodash';
+import _ from 'lodash';
 import stats from 'stats-lite';
 
 import { MixedRoute, V2Route, V3Route } from '../routers/router';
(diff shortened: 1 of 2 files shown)


src/providers/on-chain-quote-provider.ts
```diff
@@ -424,12 +430,15 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
...
+      `${this.metricsPrefix(this.chainId, useMixedRouteQuoter)}QuoteBatchSize`,
       inputs.length,
       MetricLoggerUnit.Count
     );
     metric.putMetric(
-      `${this.metricsPrefix}QuoteBatchSize_${ID_TO_NETWORK_NAME(this.chainId)}`,
+      `${this.metricsPrefix(
+        this.chainId,
+        useMixedRouteQuoter
+      )}QuoteBatchSize_${ID_TO_NETWORK_NAME(this.chainId)}`,
```
mikeki, line 441: should we store the metric prefix in a `const` and then use that `const`?

jsy1218: It might be slightly difficult, because the metrix prefix now contains the chain id, which is dynamic

PR #549, opened 2024-04-22
fix: missing chains for outside of same map config

jsy1218:
Bug fix

I was reviewing onchain quote provider config, and realized the same config map only contains Polygon. I populated for all networks to be safe for future references. Also I realized the [NETWORKS_WITH_SAME_UNISWAP_ADDRESSES](https://github.com/Uniswap/smart-order-router/blob/f73799d8025bddfea725df93f9dad6ecbe895d62/src/util/chains.ts#L53) doesn't contain all the production networks, so I need to manually populate them to be extra safe.

Add the missing chains for the same config mapping.

- **Other information**:


mikeki (approved):

10937553933:
G

PR #551, opened 2024-04-24
feat(infra): automated deploys

mr-uniswap:
This sets up a push-button deploy that uses our custom runner and token to deploy directly to NPM. 

This does not set the github tag on this repo automatically, and should probably be added later. This also does not set a deploy schedule


mikeki:
I think we might have to create a `prepublish` script in the package.json to build SOR, it's been a common problem that SOR has to be manually built before deployment.

mr-uniswap:
> I think we might have to create a `prepublish` script in the package.json to build SOR, it's been a common problem that SOR has to be manually built before deployment.

@mikeki the one nice thing with using npm ci is you rebuild it cleanly every time, so that won't be necessary unlike when you publish locally

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
+name: Release
+
+on: workflow_dispatch # TODO set this on a deploy schedule
```
mr-uniswap, line 3: Note: this is so you can "press a button and deploy" later we change this to your deploy schedule

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+          registry-url: 'https://registry.npmjs.org'
+          # Defaults to the user or organization that owns the workflow file
+          scope: '@uniswap'
+      
+      - name: Build
+        run: npm ci
```
mr-uniswap, line 37: Npm ci will give you a clean build similar to `rm -rf build node_modules && npm install && npm run build`

.github/workflows/release.yml
mikeki: Only Uniswap team members can click this button, right? :p

mikeki (approved):
LGTM let's also get Siyu's and maybe Christos opinion on the matter? but this would be way better DEVX

PR #573, opened 2024-05-13
chore(subgraph): Track filter percent

mikeki:


src/providers/v2/subgraph-provider.ts
```diff
@@ -254,6 +254,10 @@ export class V2SubgraphProvider implements IV2SubgraphProvider {
 
     metric.putMetric(`V2SubgraphProvider.chain_${this.chainId}.getPools.filter.latency`, Date.now() - beforeFilter);
     metric.putMetric(`V2SubgraphProvider.chain_${this.chainId}.getPools.filter.length`, poolsSanitized.length);
+    metric.putMetric(
+      `V2SubgraphProvider.chain_${this.chainId}.getPools.filter.percent`,
+      (poolsSanitized.length / pools.length) * 100
```
jsy1218, line 259: nit: you can specify the metric type as percent

jsy1218 (approved):

src/providers/v2/subgraph-provider.ts
mikeki: thought about it, but I think thats just a presentation thing which we dont really need much right now

PR #596, opened 2024-06-11
Fix cache batchGet keys + update unit test

xrsv:
Bug fix

We don't get fee fetch [cache hits](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#metricsV2?graph=~(metrics~(~(~'Uniswap~'TokenFeeFetcherFetchFeesFailure~'Service~'RoutingAPI)~(~'.~'TokenFeeFetcherFetchFeesSuccess~'.~'.)~(~'.~'TokenPropertiesProviderBatchGetCacheMiss~'.~'.))~sparkline~true~view~'timeSeries~stacked~false~region~'us-east-2~start~'-PT24H~end~'P0D~stat~'SampleCount~period~60)&query=~'*7bUniswap*2cService*7d*20TokenPropertiesProviderBatchGetCacheHit) due to misconfigured batchGet keys.
With this fix we should start seeing cache hits and a big qps reduction. Also updated/fixed related unit test.

- **Other information**:

### DIFF
--- a/src/providers/token-properties-provider.ts
+++ b/src/providers/token-properties-provider.ts
@@ -64,14 +64,15 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
 
     const addressesToFetchFeesOnchain: string[] = [];
     const addressesRaw = this.buildAddressesRaw(tokens);
+    const addressesCacheKeys = this.buildAddressesCacheKeys(tokens);
 
     const tokenProperties = await this.tokenPropertiesCache.batchGet(
-      addressesRaw
+      addressesCacheKeys
     );
 
     // Check if we have cached token validation results for any tokens.
     for (const address of addressesRaw) {
-      const cachedValue = tokenProperties[address];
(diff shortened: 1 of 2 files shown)


src/providers/token-properties-provider.ts
```diff
@@ -64,14 +64,15 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
 
     const addressesToFetchFeesOnchain: string[] = [];
     const addressesRaw = this.buildAddressesRaw(tokens);
```
jsy1218, line 66: Is `buildAddressesRaw` method no longer being used now? If so, I think we can delete the code block.

jsy1218 (approved):

src/providers/token-properties-provider.ts
xrsv: `buildAddressesRaw` is still used to iterate raw addresses

PR #603, opened 2024-06-12
feature: introduce a `ITokenFeeFetcher` fallback

xrsv:
feature

We currently use one `ITokenFeeFetcher` implementation to retrieve token fees

we are introducing an optional fallback `ITokenFeeFetcher` that will be used to retrieve token fee info in case primary fails, or if some tokens are not present in primary.
Plan is to use primary for our graphQL provider, then fallback to onchain fetch in case anything goes wrong or missing.

- **Other information**:
a separate PR will follow on routing-api repo

### DIFF
--- a/src/providers/token-properties-provider.ts
+++ b/src/providers/token-properties-provider.ts
@@ -44,6 +44,7 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
     private chainId: ChainId,
     private tokenPropertiesCache: ICache<TokenPropertiesResult>,
     private tokenFeeFetcher: ITokenFeeFetcher,
+    private tokenFeeFetcherFallback: ITokenFeeFetcher | undefined = undefined,
     private allowList = DEFAULT_ALLOWLIST,
     private positiveCacheEntryTTL = POSITIVE_CACHE_ENTRY_TTL,
     private negativeCacheEntryTTL = NEGATIVE_CACHE_ENTRY_TTL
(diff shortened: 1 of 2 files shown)


src/providers/token-properties-provider.ts
```diff
@@ -103,27 +105,51 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
...
+            };
+          } catch (err) {
+            log.error(
+              { err },
+              `Error fetching fees for tokens ${addressesToFetchFeesWithFallbackFetcher} using fallback`
+            );
+          }
+        }
+      }
+
```
mikeki, line 150: What do you think about creating a `GraphQlTokenFeeFetcherWithFallback` implementation of `ITokenFeeFetcher`, and we put that in `routing-api` instead of here?

That way we could avoid all changes in SOR, we just need to make sure that the `tokenFeeFetcher` can be overwritten in routing-api.

xrsv: that's not a bad idea, let me do that.

jsy1218: I like @mikeki 's idea. Then in routing-api, we can create a traffic switch fetcher class (similar to lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts), that can achieve:
1. beginning by shadow calling GQL token fetcher without affecting online path
2. compare the results between GQL token fetcher and direct on-chain token fetcher
3. gradually switching from direct on-chain token fetcher to GQL token fetcher
4. and the existing traffic switch implementation already guarantees that in case GQL token fetcher throws error, it will fallback to use GQL token fetcher.

PR #612, opened 2024-06-14
fix: check decimal diffs <= 0 and avoid throwing unchecked exceptions in gas diff token at all

jsy1218:
Bug fix

Even using DAI we can see runtime error.

We don't know why, and it shouldn't happen, because DAI and zksync ETH are both 18 decimals. But we will hotfix by checking decimaldiff <=0 so it cannot throw.

- **Other information**:


jsy1218:
* **#612** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jsy1218:
abandoned after discussing with @zhongeric

PR #634, opened 2024-07-17
fix: partially exclude fot from fee

jsy1218:
Bug fix

We exclude all the FOT output token from taking the portion.

We want to change to exclude FOT output token from taking the portion, if the FOT output token:
1. is a transfer to the external address that is not part of uniswap v2 address
2. has a double transfer FOT fee, when going through universal router custody

Both cases are handled by the updated on-chain FOT detector, so in SOR side, we just have to interweave the logic to the portion FOT exclusion part.

- **Other information**:

### DIFF
--- a/src/providers/token-fee-fetcher.ts
+++ b/src/providers/token-fee-fetcher.ts
@@ -2,8 +2,8 @@ import { BigNumber } from '@ethersproject/bignumber';
 import { BaseProvider } from '@ethersproject/providers';
 import { ChainId } from '@uniswap/sdk-core';
 
-import { TokenFeeDetector__factory } from '../types/other/factories/TokenFeeDetector__factory';
 import { TokenFeeDetector } from '../types/other/TokenFeeDetector';
+import { TokenFeeDetector__factory } from '../types/other/factories/TokenFeeDetector__factory';
 import {
   log,
   metric,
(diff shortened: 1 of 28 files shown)


jsy1218:
* **#634** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/providers/token-properties-provider.ts
```diff
@@ -56,8 +56,7 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
     const tokenToResult: TokenPropertiesMap = {};
 
     if (
-      !providerConfig?.enableFeeOnTransferFeeFetching ||
-      this.chainId !== ChainId.MAINNET
```
xrsv, line 60: this is now using GQL (with onChain fallback) - i believe GQL now supports all chains, correct?

src/providers/token-fee-fetcher.ts
```diff
@@ -27,31 +27,34 @@ type Address = string;
 export type TokenFeeResult = {
   buyFeeBps?: BigNumber;
   sellFeeBps?: BigNumber;
+  feeTakenOnTransfer?: boolean;
```
xrsv, line 30: you'll have to update GQL token fee fetcher to retrieve and expose those fields as well
see [here](https://github.com/Uniswap/routing-api/blob/19b8e7a9e591718233c056eb18a8abb211c61023/lib/graphql/graphql-token-fee-fetcher.ts#L49) and [here](https://github.com/Uniswap/routing-api/blob/19b8e7a9e591718233c056eb18a8abb211c61023/lib/graphql/graphql-queries.ts#L2)

xrsv (commented):
looks good by it probably needs some changes in routing api gql fetcher

src/providers/token-properties-provider.ts
jsy1218: exactly. it's in case GQL fallback

PR #635, opened 2024-07-19
chore(subgraph): Add error logs for subgraph pool fetching

mikeki:
# Error logs in subgraph pool fetching

In order to triage the issue of high latency when loading the subgraph, we need to know the page in which it fails.

This change will hopefully provide this information.

### DIFF
--- a/src/providers/v2/subgraph-provider.ts
+++ b/src/providers/v2/subgraph-provider.ts
@@ -162,8 +162,8 @@ export class V2SubgraphProvider implements IV2SubgraphProvider {
                   pools = [];
                   retries += 1;
                   log.info(
-                    { err },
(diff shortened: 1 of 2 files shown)


src/providers/v2/subgraph-provider.ts
```diff
@@ -162,8 +162,8 @@ export class V2SubgraphProvider implements IV2SubgraphProvider {
                   pools = [];
                   retries += 1;
                   log.info(
-                    { err },
-                    `Failed request for page of pools from subgraph. Retry attempt: ${retry}`
+                    { err, lastId },
```
jsy1218, line 165: It's log.info though.

mikeki: ohh good catch!

jsy1218 (approved):

PR #639, opened 2024-07-26
Remove unncessary mainnet check when trying ethEstimateGas

cgkol:
Remove unncessary mainnet check when trying ethEstimateGas
- Currently we go straight to tenderly for all natives except for mainnet, when we shouldnt. 

We will try ethEstimategas for all native tokens of all chains, and fallback to tenderly if it fails.
- **Other information**:


xrsv (approved):
looks good

PR #655, opened 2024-08-01
feat: implement v4 subgraph provider

jsy1218:
feature 

We don't have subgraph provider for v4

We will implement the initial subgraph provider for v4.

- **Other information**:
There's an initially deployed v4 subgraph on sepolia. We'd be able to use that subgraph for initial testing.


jsy1218:
* **#655** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cgkol (commented):
This looks almost  the same with v3, are there any differences? If not, shall we consider creating some common classes? wdyt?

jsy1218:
> This looks almost  the same with v3, are there any differences? If not, shall we consider creating some common classes? wdyt?

I actually thought about it. I can see merits in refactoring vs separation to begin with. The main benefit about separation is that if we make changes in v4 subgraph, it's easier to make changes in only a single class. One immediate thing I can see we need to change is to fetch the sqrtPriceX96 and tick, which wasn't needed in v3 subgraph.

xrsv (approved):
nice!

PR #661, opened 2024-08-15
feat: implement related v4 subgraph provider for routing-api subgraph s3 bucket download

jsy1218:
feature

We don't have caching subgraph provider, subgraph provider with fallback, and also static subgraph provider implementation for v4. 

We implement caching subgraph provider, subgraph provider with fallback, and also static subgraph provider implementation for v4.

- **Other information**:
static subgraph provider is being used by routing-api during the lambda instantiation time, as a fallback from subgraph s3 bucket download failure.


jsy1218:
* **#661** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/providers/v4/subgraph-provider.ts
```diff
@@ -17,10 +17,26 @@ export interface V4SubgraphPool {
...
+  totalValueLockedUSD: string;
+  totalValueLockedETH: string;
+  totalValueLockedUSDUntracked: string;
+};
+
 const SUBGRAPH_URL_BY_CHAIN: { [chainId in ChainId]?: string } = {
-  [ChainId.SEPOLIA]:
-    '',
-}
+  [ChainId.SEPOLIA]: '',
```
xrsv, line 38: do you have this url?

xrsv (approved):
looks good - nice refactor!
some tests need fixing but overall 👍

src/providers/v4/subgraph-provider.ts
jsy1218: We do have a alchemy private subgraph, but we cannot reveal in the public repo, because of the API key secret. This used to be ok with the hosted service of our subgraphs URL, because hosted services are free. But hosted services for Uniswap got deprecated in June, so in v4, I think we don't have any hosted service to begin with.

PR #664, opened 2024-08-19
feat: implement v4 candidate pools from top tvl v4 subgraph

jsy1218:

- **Other information**:


jsy1218:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/664?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#664** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#661** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/661?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/providers/subgraph-provider.ts
```diff
@@ -22,7 +20,41 @@ export interface ISubgraphProvider<TSubgraphPool extends SubgraphPool> {
...
+    symbol: string;
+    id: string;
+  };
+  totalValueLockedUSD: string;
+  totalValueLockedETH: string;
+  totalValueLockedUSDUntracked: string;
+};
+
+export abstract class SubgraphProvider<
+  TRawSubgraphPool extends V3V4RawSubgraphPool,
```
cgkol, line 55: is this the right way to go about it? because technically subgraph pool can also be from v2 right?

cgkol: i understand its a replcament but curious why we did not include v2 in this subgraphProvider originally

src/routers/alpha-router/config.ts
```diff
@@ -30,6 +30,14 @@ export const DEFAULT_ROUTING_CONFIG_BY_CHAIN = (
           topNWithEachBaseToken: 3,
           topNWithBaseToken: 3,
         },
+        v4PoolSelection: {
+          topN: 2,
```
cgkol, line 34: i assume we are following v3?  how come v2 different?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1955,6 +1988,12 @@ export class AlphaRouter
 
     const quotePromises: Promise<GetQuotesResult>[] = [];
 
+    // for v4, for now we explicitly require people to specify
+    if (v4ProtocolSpecified) {
```
cgkol, line 1992: shouldnt we do this before the rpoise? otherwise we are just wasting resources right?

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -365,6 +380,450 @@ function findCrossProtocolMissingPools<
   return selectedPools;
 }
 
+export type V4CandidatePools = {
+  poolAccessor: V4PoolAccessor;
+  candidatePools: CandidatePoolsBySelectionCriteria;
+  subgraphPools: V4SubgraphPool[];
+};
+
+export async function getV4CandidatePools({
```
cgkol, line 389: I see very few differences between v4 and v3 (tick sizing, naming + one more point in the middle). Shall we try to collapse the functions?

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -365,6 +380,450 @@ function findCrossProtocolMissingPools<
...
+        (subgraphPool.token0.id == tokenOutAddress ||
+          subgraphPool.token1.id == tokenOutAddress)
+      );
+    })
+    .slice(0, topNTokenInOut)
+    .value();
+
+  addToAddressSet(topByTVLUsingTokenOut);
+
+  const topByTVLUsingTokenInSecondHops = _(topByTVLUsingTokenIn)
```
cgkol, line 647: also for this one, now we got to make sure we use both v2 and v3 as par of top from second hops right?

src/providers/subgraph-provider.ts
jsy1218: Because I think it's not worth the effort to refactor v2 subgraph. V3 and V4 subgraph pools have similar data structures like ticks, fee tiers, etc., meanwhile V2 goes with reserves. In future, we are more likely to stay with ticks, fee tiers, etc., for concentrated liquidities, which leaves V2 a special case.

src/routers/alpha-router/functions/get-candidate-pools.ts
jsy1218: I'm hesitant to refactor in this PR, due to the large function. I'd be more confident to refactor V3CandidatePools against V4CandidatePools when we add more testing coverages against V3CandidatePools. Also, I don't see we will have large amount of v4 subgraph pools for a while, once we launch the v4 routing, which leaves us some time to address this tech debt.

I'm thinking to add a tech debt ticket for this. lmk what you think.

src/routers/alpha-router/functions/get-candidate-pools.ts
jsy1218: actually function `getV4CandidatePools` should only return v4 subgraph pools, regardless which heuristics the pool would satisfy.  We have a separate function `getMixedRouteCandidatePools` that will return a combination of v2, v3, v4 pools. Also the mixed cross liquidity pools heuristic (that can include wstETH -> DOGE) is also in the `getMixedRouteCandidatePools`

src/routers/alpha-router/alpha-router.ts
jsy1218: I think I didn't follow. Maybe we can chat more offline.

jsy1218: My understanding is that the upstream dependencies will always specify a protocol version when they request for a route. I think this if block shouldn't get entered, since v4 is not supported yet. Not sure if answers?

src/routers/alpha-router/functions/get-candidate-pools.ts
cgkol: can yoou add the ticket? then you can resolve the comment:)

src/routers/alpha-router/functions/get-candidate-pools.ts
cgkol: oh woops, got confused my bad;

cgkol (approved):
✅

src/routers/alpha-router/functions/get-candidate-pools.ts
jsy1218: Done! https://linear.app/uniswap/issue/ROUTE-241/tech-debt-refactor-getv3candidatepools-against-getv4candidatepools

src/routers/alpha-router/config.ts
jsy1218: I think prior to launch Blast, we expected V2 pools to be more popular on Blast, because Blast stablecoin USDB and native ETH is a yield-bearing token, which means it works well on Uniswap V2 pools, but not v3 pools. Hence the search space on Blast v2 pools will be slightly larger than Blast V3 pools.

Blast is kinda unique case, since we don't see other Uniswap supported chains having yield-bearing native token and stablecoin.

src/routers/alpha-router/functions/get-candidate-pools.ts
```diff
@@ -365,6 +380,451 @@ function findCrossProtocolMissingPools<
...
+    .value();
+
+  const tokenAddresses = _(subgraphPools)
+    .flatMap((subgraphPool) => [subgraphPool.token0.id, subgraphPool.token1.id])
+    .compact()
+    .uniq()
+    .value();
+
+  log.info(
+    `Getting the ${tokenAddresses.length} tokens within the ${subgraphPools.length} V3 pools we are considering`
```
xrsv, line 724: change to `V4 pools we are considering`

src/providers/v4/subgraph-provider.ts
```diff
@@ -21,6 +23,8 @@ export interface V4SubgraphPool {
 export type V4RawSubgraphPool = {
   id: string;
   feeTier: string;
+  tickSpacing: string; // TODO: waiting on v4 subgraph to return tickSpacing
+  hooks: string;
```
xrsv, line 27: curious how this `hooks` value will look like?

src/providers/subgraph-provider.ts
```diff
@@ -22,7 +20,41 @@ export interface ISubgraphProvider<TSubgraphPool extends SubgraphPool> {
...
+  TSubgraphPool extends V3V4SubgraphPool
```
xrsv, line 56: got a bit confused with `V3V4SubgraphPool` vs `V3SubgraphPool`,`V3SubgraphPool` (those are still defined in v3/&v4/ subgraph-provider.ts). so here you are saying that both should have those common fields?

xrsv (approved):
looks good - left some comment for clarification

src/providers/subgraph-provider.ts
jsy1218: yep, v3 v4 have a lot of common fields, meanwhile v2 is not the case.

jsy1218: sadly its kinda confusing. because v3 v4 have concentrated liquidities, so v3 v4 pools data structure look very similar, meanwhile v2 pool data structure is entirely different.

In future, I see us stay in concentrated liquidities, hence I see the effort to refactor v2 subgraph provider is not best spent.

src/providers/v4/subgraph-provider.ts
jsy1218: aw hooks address will be mined prior to its deployment along with the pool. Uniswap Foundation has the [HookMiner](https://github.com/uniswapfoundation/v4-template/blob/b47e8f622852ea4bebbc0afb131d0c9458ecb8a7/test/utils/HookMiner.sol#L4) implementation. Hook devs don't have to use this HookMiner, but the `hooks` will follow certain pattern.

jsy1218:
### Merge activity

* **Aug 21, 5:12 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/smart-order-router/664).

PR #673, opened 2024-08-23
fix: ticks crossing gas estimate

jsy1218:
Bug fix

We have incorrect tick crossing gas estimate logic

We are fixing tick crossing gas estimate logic. This is relevant for v3 and mixed (v3 part of mixed)

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/gas-models/gas-model.ts
+++ b/src/routers/alpha-router/gas-models/gas-model.ts
@@ -243,6 +243,22 @@ export abstract class IOnChainGasModelFactory {
   }: BuildOnChainGasModelFactoryType): Promise<
     IGasModel<V3RouteWithValidQuote | MixedRouteWithValidQuote>
   >;
+
+  protected totalInitializedTicksCrossed(
+    initializedTicksCrossedList: number[]
+  ) {
+    let ticksCrossed = 0;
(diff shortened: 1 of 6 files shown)


jsy1218:
* **#673** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cgkol (commented):
Shall we add some tests? also in prod, how will we verify that it works?

xrsv:
curious how did we figure out that gas estimation logic was incorrect? 
do you think division by zero error could be related?

jsy1218:
> Shall we add some tests? also in prod, how will we verify that it works?
> 

yes on adding tests. I do think it's a bit difficult to verify it's fully working in prod, because SOR gas estimate has been knowingly off. Better way is to let the quoter returns a somewhat accurate gas estimate, and SOR just uses it in the gas model classes.

jsy1218:
> curious how did we figure out that gas estimation logic was incorrect? 
> do you think division by zero error could be related?

@xrsv and I chatted offline. It's a combination of discussing with protocol team, as well as reading the pool swap contract. Division by zero is prob due to some other reason, my guess is because the quote amount is too small getting close to zero.

xrsv (approved):

PR #677, opened 2024-08-26
fix: for token pair only quote against v2 quoter

jsy1218:
Bug fix

We will attempt to quote a FOT pair against v3 quoter or mixed quoter.

FOT doesn't work in v3 pool. This means quoting against v3 quoter or mixed quoter (which goes through at least one v3 pool of the mixed route) doesn't work. If we see the tokenIn is sell FOT tax or tokenOut is buy FOT tax, then we only go through v2 quoter.

- **Other information**:
Testing wise, I'm relying on the existing e2e tests in both SOR and routing-api. On SOR level, it never goes through cached routes, so it will test against `swapRouteFromChain`. On routing-api level, we constantly run the code pipeline, so that the same FOT e2e test will likely go through the cached routes, i.e. `swapRouteFromCache`.

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -130,7 +130,8 @@ import {
   SwapToRatioStatus,
   SwapType,
   V2Route,
-  V3Route, V4Route
+  V3Route,
+  V4Route,
 } from '../router';
 
 import { CachingV4PoolProvider } from '../../providers/v4/caching-pool-provider';
@@ -1178,8 +1179,14 @@ export class AlphaRouter
     // We want to log the fee on transfer output tokens that we are taking fee or not
     // Ideally the trade size (normalized in USD) would be ideal to log here, but we don't have spot price of output tokens here.
     // We have to make sure token out is FOT with either buy/sell fee bps > 0
-    if (tokenOutProperties[tokenOut.address.toLowerCase()]?.tokenFeeResult?.buyFeeBps?.gt(0) ||
-        tokenOutProperties[tokenOut.address.toLowerCase()]?.tokenFeeResult?.sellFeeBps?.gt(0)) {
+    if (
+      tokenOutProperties[
+        tokenOut.address.toLowerCase()
+      ]?.tokenFeeResult?.buyFeeBps?.gt(0) ||
+      tokenOutProperties[
+        tokenOut.address.toLowerCase()
+      ]?.tokenFeeResult?.sellFeeBps?.gt(0)
+    ) {
       if (feeTakenOnTransfer || externalTransferFailed) {
         // also to be extra safe, in case of FOT with feeTakenOnTransfer or externalTransferFailed,
         // we nullify the fee and flat fee to avoid any potential issues.
@@ -1390,6 +1397,8 @@ export class AlphaRouter
       Promise.resolve(null);
     if (cachedRoutes) {
       swapRouteFromCachePromise = this.getSwapRouteFromCache(
+        tokenIn,
+        tokenOut,
         cachedRoutes,
         await blockNumber,
         amount,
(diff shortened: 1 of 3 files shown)


jsy1218:
* **#677** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/677?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2175,58 +2230,60 @@ export class AlphaRouter
...
-    // Maybe Quote mixed routes
-    // if MixedProtocol is specified or no protocol is specified and v2 is supported AND tradeType is ExactIn
-    // AND is Mainnet or Gorli
-    if (shouldQueryMixedProtocol && mixedProtocolAllowed) {
-      log.info({ protocols, tradeType }, 'Routing across MixedRoutes');
+    if (!fotInDirectSwap) {
+      // Maybe Quote mixed routes
+      // if MixedProtocol is specified or no protocol is specified and v2 is supported AND tradeType is ExactIn
+      // AND is Mainnet or Gorli
+      if (shouldQueryMixedProtocol && mixedProtocolAllowed) {
```
cgkol, line 2237: shall you add it in this if statement?+

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1733,6 +1748,22 @@ export class AlphaRouter
     swapConfig?: SwapOptions,
     providerConfig?: ProviderConfig
   ): Promise<BestSwapRoute | null> {
+    const tokenPairProperties =
```
cgkol, line 1751: nit: maybe lets make this a util function? seems like we call it a few times?

cgkol (commented):
any tests we might want to add here to make sure this deos not break? i would suggest something like confirming that with fot buy ro sell we only get v2 candidate pools (good way to also proof v4 and future versions as well)

jsy1218:
> any tests we might want to add here to make sure this deos not break? i would suggest something like confirming that with fot buy ro sell we only get v2 candidate pools (good way to also proof v4 and future versions as well)

potentially we can add tests in alpha-router.test.ts.

src/routers/alpha-router/alpha-router.ts
jsy1218: I think it's more readable to make a outer-scoped if statement for `fotInDirectSwap`.  Also there are some complex conditionals that makes it prone to make coding mistakes, e.g. 

```
if (
        v3ProtocolSpecified ||
        noProtocolsSpecified ||
        (shouldQueryMixedProtocol && mixedProtocolAllowed)
   ) {}
```

I can make it such that:

```
if (
        (v3ProtocolSpecified ||
        noProtocolsSpecified ||
        (shouldQueryMixedProtocol && mixedProtocolAllowed)) &&
       fotInDirectSwap
   ) {}

```

But it's more readable with:

```
if (!fotInDirectSwap) {
      if (
        v3ProtocolSpecified ||
        noProtocolsSpecified ||
        (shouldQueryMixedProtocol && mixedProtocolAllowed)
      ) {
       }
}
```

src/routers/alpha-router/alpha-router.ts
jsy1218: I tried it, and somehow some unit tests in `alpha-router.test.ts` started failing.

jsy1218: I suspect it's mocking the token properties in those unit tests at the alpha-router class, which makes it kinda hard to do this kind of refactoring.

src/routers/alpha-router/alpha-router.ts
cgkol: fair enough

PR #686, opened 2024-09-04
feat: exclude certain protocol versions from mixed routing

jsy1218:
feature

For mixed protocol version request, we don't have the ability to exclude a certain route with a certain pool version. Some clients would love to still request mixed routes, but they don't want any of the mixed route to go through any v4 pools.

We will expose a new additional generic field `excludedProtocolsFromMixed` to be able to exclude certain protocol version pool routes from the mixed routes.

- **Other information**:
We can be somewhat limited in scope in the protocol exclusions here. If someone says I don't want pure v3-pool routes returned from SOR, then they simply don't need to request `v3` in the request protocol versions array. Likewise for any other version specific protocol request. This eases the implementation complexity, meanwhile we still have a bit of flexibility, in case in future client wants to exclude other protocol version from mixed routing.

Also from routing-api perspective, it will accept a new query string param called 'mixedExcludeV4' and hardcode `excludedProtocolsFromMixed` to be `[V4]`, since at routing-api layer it can be more opinionated.

Once we implement mixed routing with v4 support in SOR, we can add additional alpha router integ tests to verify.

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -87,6 +87,7 @@ import {
 import { IV3SubgraphProvider } from '../../providers/v3/subgraph-provider';
 import { Erc20__factory } from '../../types/other/factories/Erc20__factory';
 import {
+  shouldWipeoutCachedRoutes,
   SWAP_ROUTER_02_ADDRESSES,
   V4_SUPPORTED,
   WRAPPED_NATIVE_CURRENCY
@@ -165,7 +166,7 @@ import {
   SubgraphPool,
   V2CandidatePools,
   V3CandidatePools,
-  V4CandidatePools,
+  V4CandidatePools
 } from './functions/get-candidate-pools';
 import { NATIVE_OVERHEAD } from './gas-models/gas-costs';
 import {
@@ -399,6 +400,10 @@ export type AlphaRouterConfig = {
    * will be used.
    */
   protocols?: Protocol[];
+  /**
+   * The protocols-version pools to be excluded from the mixed routes.
+   */
+  excludedProtocolsFromMixed?: Protocol[];
   /**
    * Config for selecting which pools to consider routing via on V2.
    */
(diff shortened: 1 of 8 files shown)


jsy1218:
* **#686** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1356,6 +1361,15 @@ export class AlphaRouter
       );
     }
 
+    if (
```
xrsv, line 1364: should we not even fetch them in this case instead of fetching and then setting to null?

xrsv: also does that mean that we won't use cached routes at all in this case?
would that impact latency?

jsy1218: we should still fetch cached routes for mixed protocol request. the reason is because there's a chance that cached routes do not contain any excluded protocol-versioned pool in any of its cached routes. We should only nullify the cached routes, when any of the cached routes is a mixed route, and such mixed route contains a protocol-versioned pool, that should be excluded.

Although I did realize then, that we should only nullify the cached routes, only when the intent=caching, but not when intent=quote. intent=quote means it's an online request from routing-api quote endpoint perspective. If it's a cache hit, we still want routing-api to return the ineligible cached route, if it contains a mixed route, such that the mixed route contains a protocol-versioned pool that should be excluded. This is during intent=quote. For every intent=quote, routing-api self invokes its own lambda again via intent=caching. This is when we should nullify the cached routes, and re-compute the routes, if any of any of the cached routes is a mixed route, and such mixed route contains a protocol-versioned pool, that should be excluded.

This seems like a good tradeoff between online quote endpoint perf vs the correctness returned from the quote endpoint for the first quote request. In reality, because the top route (that includes the mixed route) is so low as compared to a pure v2 or v3 route, I think this tradeoff of favoring endpoint perf over first quote response correctness is warranted:

Screenshot 2024-09-05 at 4 28 58 PM

I will make a change to only nullify the cached routes during intent=caching

jsy1218: done

xrsv (approved):
looks reasonable - would be good to keep an eye on latency after checkin

jsy1218:
### Merge activity

* **Sep 5, 10:31 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/smart-order-router/686).

PR #702, opened 2024-09-17
feat: cached routes to support native currencies

jsy1218:
feature

cached routes do not support native currency routing

we want cached routes to support native currency routing. This is import for v4 native currency routing support.

- **Other information**:
Although we changed the method signature to accept Currency, for now we still pass in Tokens for tokenIn and tokenOut, so that this change should still remain no-op for now.

### DIFF
--- a/src/util/addresses.ts
+++ b/src/util/addresses.ts
@@ -294,9 +294,17 @@ export const WETH9: {
 export const BEACON_CHAIN_DEPOSIT_ADDRESS =
   '0x00000000219ab540356cBB839Cbe05303d7705Fa';
 
-export function getAddress(currency: Currency): string {
+export function getAddressLowerCase(currency: Currency): string {
   if (currency.isToken) {
     return currency.address.toLowerCase();
+  } else {
+    return ADDRESS_ZERO;
+  }
+}
+
+export function getAddress(currency: Currency): string {
+  if (currency.isToken) {
+    return currency.address;
   } else {
     return ADDRESS_ZERO;
   }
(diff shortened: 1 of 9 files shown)


jsy1218:
* **#702** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/702?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/util/addresses.ts
```diff
@@ -294,10 +294,18 @@ export const WETH9: {
...
-export function getAddress(currency: Currency): string {
+export function getAddressLowerCase(currency: Currency): string {
   if (currency.isToken) {
     return currency.address.toLowerCase();
   } else {
     return ADDRESS_ZERO;
   }
 }
+
+export function getAddress(currency: Currency): string {
```
xrsv, line 305: what's the reason to not always use the lowercase?

xrsv (approved):
looks good - left one comment about lowercase vs not method.

src/util/addresses.ts
jsy1218: so in case of cached routes, we aren't lower casing the token addresses in routing-api (https://github.com/Uniswap/routing-api/blob/main/lib/handlers/marshalling/token-marshaller.ts#L18). This means the cached routes key in dynamo remains the original form. If we start lower casing the cached routes token addresses, we will have lots of cached routes cache misses for a short period of time.

jsy1218:
### Merge activity

* **Sep 17, 9:34 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/smart-order-router/702).

PR #715, opened 2024-09-22
feat(breaking): top-level alpha router native currencyIn, currencyIn, quoteCurrency support

jsy1218:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
breaking feature

- **What is the current behavior?** (You can also link to an open issue here)
Top-level alpha router doesn't support native currency routing.

- **What is the new behavior (if this is a feature change)?**
Top-level alpha router should support native currency routing.

- **Other information**:
breaking because I changed the method name in CachedRoute class. Will have to major version bump SOR.

PR #723, opened 2024-09-30
feat: part 1 native currency routing support by removing .wrapped in implementation classes

jsy1218:
feature

there are several code places where we simply wrap the currency, and thus not able to support native currency routing in v4.

we are removing the .wrapped implementations. We keep the alpha router to pass in the wrapped tokens into the quoters in this PR, for backward compatibility.

- **Other information**:
re-do https://github.com/Uniswap/smart-order-router/pull/715 with breakdowns

### DIFF
--- a/src/providers/caching/route/model/cached-route.ts
+++ b/src/providers/caching/route/model/cached-route.ts
@@ -1,5 +1,5 @@
 import { Protocol } from '@uniswap/router-sdk';
-import { Token } from '@uniswap/sdk-core';
+import { Currency } from '@uniswap/sdk-core';
 import { Pair } from '@uniswap/v2-sdk';
 import { Pool as V3Pool } from '@uniswap/v3-sdk';
 import { Pool as V4Pool } from '@uniswap/v4-sdk';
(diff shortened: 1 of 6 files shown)


jsy1218:
* **#723** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/723?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/providers/caching/route/model/cached-route.ts
```diff
@@ -44,24 +45,21 @@ export class CachedRoute<Route extends SupportedRoutes> {
     return this.route.protocol;
   }
 
-  // TODO: ROUTE-217 - Support native currency routing in V4
-  public get tokenIn(): Token {
-    return this.route.input.wrapped;
+  public get currencyIn(): Currency {
+    return this.route.input;
```
cgkol, line 49: this affected v3 and v4 caching though right? for v3 we want the same and for v4 we want something different right?

cgkol (approved):

src/providers/caching/route/model/cached-route.ts
jsy1218: I think it's ok. The reason is because type Currency is a union type of Token and NativeCurrency, which means it's backward compatible with v2/v3/mixed cached routes input and output type of Token, meanwhile v4 cached routes can have either a Token or NativeCurrency token.

PR #746, opened 2024-10-11
chore: bump sor to 4.7.0 - feat: SOR level implementation to support 1) expanding mixed quoter v1 to L2s 2) support migrate mixed quoter v1 to mixed quoter v2 across chains

jsy1218:
feature

right now, mixed routing is only supported on L1 (plus L1 testnet).

we want to be able to expand mixed routing support to L2s. There are multiple stages:
1) routing-api to shadow sample mixed routing against mixed quoter v1 across L2s. There's no accuracy to compare, just make sure routing-api shadow mixed routing success rate is high - existing metrics in routing-api `ON_CHAIN_QUOTE_PROVIDER_${this.EXACT_IN_METRIC}_TRAFFIC_TARGET_FAILED_CHAIN_ID_${this.chainId}`
2) routing-api to live traffic switch mixed routing against mixed quoter v1  across L2s. Again there's no accuracy to compare, but we can live switch gradually per chain. We will check the quote endpoint success rate doesn;'t drop.
3) routing-api to shadow sampling mixed routing against mixed quoter v2. mixed quoter v2 contains v4 pools, but v4 is only deployed to sepolia, hence this can only happen after we deployed v4 to production nets. There are accuracy comparison.
4) routing-api to live traffic switch mixed routing against mixed quoter v2. We will check the quote endpoint success rate doesn;'t drop.

- **Other information**:
routing-api needs some changes as well, in order to achieve 1-4

### DIFF
--- a/src/providers/on-chain-quote-provider.ts
+++ b/src/providers/on-chain-quote-provider.ts
@@ -349,30 +349,35 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
     },
     protected batchParams: (
       optimisticCachedRoutes: boolean,
-      useMixedRouteQuoter: boolean,
-      mixedRouteContainsV4Pool: boolean,
       protocol: Protocol
-    ) => BatchParams = (
-      _optimisticCachedRoutes,
-      _useMixedRouteQuoter,
-      _mixedRouteContainsV4Pool,
-      _protocol
-    ) => {
+    ) => BatchParams = (_optimisticCachedRoutes, _protocol) => {
       return {
         multicallChunk: 150,
         gasLimitPerCall: 1_000_000,
         quoteMinSuccessRate: 0.2,
       };
     },
-    protected gasErrorFailureOverride: FailureOverrides = {
-      gasLimitOverride: 1_500_000,
-      multicallChunk: 100,
+    protected gasErrorFailureOverride: (
+      protocol: Protocol
+    ) => FailureOverrides = (_protocol: Protocol) => {
+      return {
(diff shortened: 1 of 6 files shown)


jsy1218:
* **#746** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xrsv (approved):
looks good

src/providers/on-chain-quote-provider.ts
```diff
@@ -611,6 +633,12 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
             (route as MixedRoute).pools.some((pool) => pool instanceof V4Pool)
         )
       : false;
+    const protocol = useMixedRouteQuoter
+      ? Protocol.MIXED
```
cgkol, line 637: this is so hard to read, can we just do if else? 😅

cgkol (approved):
This PR is a no-op

src/providers/on-chain-quote-provider.ts
jsy1218: i think i was using ternary instead of if-else for conciseness.. because otherwise, it would look like the following:

```
let protocol = undefined;

    if (useMixedRouteQuoter) {
      protocol = Protocol.MIXED;
    } else {
      protocol = useV4RouteQuoter ? Protocol.V4 : Protocol
    }
```

i wanted to avoid protocol being a mutable variable.

PR #747, opened 2024-10-11
chore: bump sor to 4.7.5 - fix: cached routes cache invalidation

jsy1218:
Bug fix

if the cached routes is invalid, meaning it cant get any swap route at all, it does not get cache invalidated.

we will invalidate the cached routes by checking to see if the swap route from cache is invalid

- **Other information**:
we have to roll out with percentage. because there's a chance there are many invalid cached routes in prod.

This is somewhat critical to fix prior to V4, because there's a chance that routing caches invalid v4-pool routes. Previously this wasn't much of a concern,. since V3 launched in 2021, and cached routes was rolled out in 2023, so the v2 and v3 pool TVLs won't change that much, although concentrated liquidities can change based on the market pricing.

I expect with v4, there will be more dynamics from 0 TVL to some TVLs, hence cached routes have more chance of caching invalid v4-pool routes.

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1795,7 +1795,7 @@ export class AlphaRouter
         // in AWS metrics, one can investigate, by:
         // 1) seeing the overall metrics count of SetCachedRoute_NewPath and SetCachedRoute_OldPath. SetCachedRoute_NewPath should steadily go up, while SetCachedRoute_OldPath should go down.
         // 2) using the same requestId, one should see eventually when SetCachedRoute_NewPath metric is logged, SetCachedRoute_OldPath metric should not be called.
-        metric.putMetric(`SetCachedRoute_OldPath`, 1, MetricLoggerUnit.Count);
+        metric.putMetric(`SetCachedRoute_OldPath_INTENT_${routingConfig.intent}`, 1, MetricLoggerUnit.Count);
       }
 
       // Generate the object to be cached
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -10,7 +10,6 @@ import {
   TradeType,
 } from '@uniswap/sdk-core';
 import { TokenList } from '@uniswap/token-lists';
-import { UniversalRouterVersion } from '@uniswap/universal-router-sdk';
 import { Pool, Position, SqrtPriceMath, TickMath } from '@uniswap/v3-sdk';
 import retry from 'async-retry';
 import JSBI from 'jsbi';
@@ -149,6 +148,7 @@ import {
   V4Route,
 } from '../router';
 
+import { UniversalRouterVersion } from '@uniswap/universal-router-sdk';
 import { INTENT } from '../../util/intent';
 import {
   DEFAULT_ROUTING_CONFIG_BY_CHAIN,
@@ -324,11 +324,6 @@ export type AlphaRouterParams = {
    * All the supported mixed chains configuration
    */
   mixedSupported?: ChainId[];
-
-  /**
-   * The version of the universal router to use.
-   */
-  universalRouterVersion?: UniversalRouterVersion;
 
   /**
    * The v4 pool params to be used for the v4 pool provider.
@@ -566,7 +561,6 @@ export class AlphaRouter
   protected v2Supported?: ChainId[];
   protected v4Supported?: ChainId[];
   protected mixedSupported?: ChainId[];
-  protected universalRouterVersion?: UniversalRouterVersion;
   protected v4PoolParams?: Array<[number, number, string]>;
   protected cachedRoutesCacheInvalidationFixRolloutPercentage?: number;
 
@@ -599,7 +593,6 @@ export class AlphaRouter
     v2Supported,
     v4Supported,
     mixedSupported,
-    universalRouterVersion,
     v4PoolParams,
     cachedRoutesCacheInvalidationFixRolloutPercentage,
   }: AlphaRouterParams) {
@@ -1065,8 +1058,6 @@ export class AlphaRouter
     this.v2Supported = v2Supported ?? V2_SUPPORTED;
     this.v4Supported = v4Supported ?? V4_SUPPORTED;
     this.mixedSupported = mixedSupported ?? MIXED_SUPPORTED;
-    this.universalRouterVersion =
-      universalRouterVersion ?? UniversalRouterVersion.V1_2;
 
     this.cachedRoutesCacheInvalidationFixRolloutPercentage =
       cachedRoutesCacheInvalidationFixRolloutPercentage;
@@ -1795,7 +1786,11 @@ export class AlphaRouter
         // in AWS metrics, one can investigate, by:
         // 1) seeing the overall metrics count of SetCachedRoute_NewPath and SetCachedRoute_OldPath. SetCachedRoute_NewPath should steadily go up, while SetCachedRoute_OldPath should go down.
         // 2) using the same requestId, one should see eventually when SetCachedRoute_NewPath metric is logged, SetCachedRoute_OldPath metric should not be called.
-        metric.putMetric(`SetCachedRoute_OldPath_INTENT_${routingConfig.intent}`, 1, MetricLoggerUnit.Count);
+        metric.putMetric(
+          `SetCachedRoute_OldPath_INTENT_${routingConfig.intent}`,
+          1,
+          MetricLoggerUnit.Count
+        );
       }
 
       // Generate the object to be cached
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -514,11 +514,6 @@ export type AlphaRouterConfig = {
    */
   universalRouterVersion?: UniversalRouterVersion;
   /**
-   * We want to rollout the cached routes cache invalidation carefully.
-   * Because a wrong fix might impact prod success rate and/or latency.
-   */
-  cachedRoutesCacheInvalidationFixRolloutPercentage?: number;
-  /**
    * pass in routing-api intent to align the intent between routing-api and SOR
    */
   intent?: INTENT;
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1741,7 +1741,12 @@ export class AlphaRouter
     // because we know cached routes in prod dont filter by blocks-to-live
     // so that we know that swapRouteFromChain is always not populated, because
     // if (!cachedRoutes || cacheMode !== CacheMode.Livemode) above always have the cachedRoutes as populated
-    if (this.routeCachingProvider && swapRouteFromChain) {
+    if (
+      this.routeCachingProvider &&
+      routingConfig.writeToCachedRoutes &&
+      cacheMode !== CacheMode.Darkmode &&
+      swapRouteFromChain
+    ) {
       if (newSetCachedRoutesPath) {
         // SetCachedRoute_NewPath and SetCachedRoute_OldPath metrics might have counts during short timeframe.
         // over time, we should expect to see less SetCachedRoute_OldPath metrics count.
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -334,6 +334,12 @@ export type AlphaRouterParams = {
    * fee tiers, tickspacings, and hook addresses
    */
   v4PoolParams?: Array<[number, number, string]>;
+
+  /**
+   * We want to rollout the cached routes cache invalidation carefully.
+   * Because a wrong fix might impact prod success rate and/or latency.
+   */
+  cachedRoutesCacheInvalidationFixRolloutPercentage?: number;
 };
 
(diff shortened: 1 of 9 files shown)


jsy1218:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/747?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#747** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#746** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

jsy1218:
i will add a test in alpha-router.integ-test. i have an idea how to repro this within SOR. this kind of test setup is first time, so it may take a bit of time.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1631,6 +1642,86 @@ export class AlphaRouter
...
+        (this.cachedRoutesCacheInvalidationFixRolloutPercentage ?? 0)
+      ) {
+        // there's a possibility that cachedRoutes object is not null/undefined, but it could cause swapRouteFromCache to be null/undefined
+        // because such cached routes already becomes invalid.
+        // in that case then, we want to try swapRouteFromChain.
+        // If CachedRoutes is populated, then swapRouteFromChain is not being invoked at all in
+        // https://github.com/Uniswap/smart-order-router/blob/66927c4e4f7e75138df48afab1c33a6c6d8d8f15/src/routers/alpha-router/alpha-router.ts#L1469
+        if (
+          this.routeCachingProvider &&
+          routingConfig.writeToCachedRoutes &&
```
cgkol, line 1660: even if `writeToCachedRoutes` is false, dont we want to do `getSwapRouteFromChain` if`swapRouteFromCache` is invalid/null?

cgkol: i guess by definition no, but `writeToCachedRoutes` is nowadays always set to true right?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1631,6 +1642,86 @@ export class AlphaRouter
...
+          cacheMode !== CacheMode.Darkmode &&
+          // if cachedRoutes is defined, then swapRouteFromChain cannot be defined
```
cgkol, line 1662: is this because of the if logic at statement at line 1504?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1631,6 +1642,86 @@ export class AlphaRouter
...
+          // there's a possibility that swapRouteFromCache is undefined, in cached routes become invalid
+          // in that case, we want to make sure we swap from chain.
+          // we use optimisticCachedRoutes = false to make sure it's intent=caching path
+          cachedRoutes &&
```
cgkol, line 1666: so this is the chance where the entry does exist, but the entry itself is null right? because `swapRouteFromCache` will by definition get created if `cachedRoutes` is true right?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1662,45 +1756,14 @@ export class AlphaRouter
         amount.toExact()
       );
 
-      if (routesToCache) {
```
cgkol, line 1665: you still want to check that `routesToCache` exists right? if its null we should not even add it. Also, i dont understand how we can end on the state we are looking for above, if we always check that `routesToCache` is not null?

cgkol: oh woops missed that my bad

src/routers/alpha-router/alpha-router.ts
jsy1218: aw i see.. i don't see that's possible. becauses during the cache set in routing-api, individual route has to exist with the route id in https://github.com/Uniswap/routing-api/blob/main/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L388, we can sync more offline

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1645,6 +1652,105 @@ export class AlphaRouter
...
+    // as long as it's within 24 hours sliding window TTL
+    if (shouldEnableCachedRoutesCacheInvalidationFix) {
+      // theoretically, when routingConfig.intent === INTENT.CACHING, optimisticCachedRoutes should be false
+      // so that we can always pass in cachedRoutes?.notExpired(await blockNumber, !routingConfig.optimisticCachedRoutes)
+      // but just to be safe, we just hardcode true when checking the cached routes expiry for write update
+      // we decide to not check cached routes expiry in the read path anyway
+      if (!cachedRoutes?.notExpired(await blockNumber, true)) {
+        // optimisticCachedRoutes === false means at routing-api level, we only want to set cached routes during intent=caching, not intent=quote
+        // this means during the online quote endpoint path, we should not reset cached routes
+        if (routingConfig.intent === INTENT.CACHING) {
```
xrsv, line 1672: how often/ when is INTENT.CACHING sent?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -324,16 +325,17 @@ export type AlphaRouterParams = {
    */
   mixedSupported?: ChainId[];
 
-  /**
-   * The version of the universal router to use.
-   */
-  universalRouterVersion?: UniversalRouterVersion;
```
xrsv, line 330: not used anymore? curious why this was introduced originally?

xrsv (commented):
thanks - left few comments

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1645,6 +1652,105 @@ export class AlphaRouter
...
+                v3GasModel,
+                v4GasModel,
+                mixedRouteGasModel,
+                gasPriceWei,
+                v2GasModel,
+                swapConfig,
+                providerConfig
+              );
+
+          // intentionally a fire and forget, because getSwapRouteFromChain can be expensive
```
cgkol, line 1701: this only sets new one if it needs to update, but actually does not delete old bad ones right?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1645,6 +1652,105 @@ export class AlphaRouter
...
+          swapRouteFromChainAgainPromise.then(async (swapRouteFromChain) => {
+            if (swapRouteFromChain) {
```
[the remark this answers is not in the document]
cgkol: so we are not actually fixing the bug

src/routers/alpha-router/alpha-router.ts
jsy1218: i introduced duplicate, at both [ctor](https://github.com/Uniswap/smart-order-router/blob/b3b8555a781a6373bf943a9cccc8696d121f4a8d/src/routers/alpha-router/alpha-router.ts#L330) level and method [param](https://github.com/Uniswap/smart-order-router/blob/b3b8555a781a6373bf943a9cccc8696d121f4a8d/src/routers/alpha-router/alpha-router.ts#L513) level. it's only needed at method param level, so im cleaning up in the ctor level

src/routers/alpha-router/alpha-router.ts
jsy1218: so every quote request (intent = quote) from upstream will have a corresponding routing lambda async caching https://github.com/Uniswap/routing-api/blob/08646b11ca1246ba4c666f4909661e234520a1ee/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L346.

src/routers/alpha-router/alpha-router.ts
jsy1218: i mean its an in-place update in dynamo, because the partition key doesnt change, its still https://github.com/Uniswap/routing-api/blob/08646b11ca1246ba4c666f4909661e234520a1ee/lib/handlers/router-entities/route-caching/model/pair-trade-type-chain-id.ts#L29

`${this.currencyIn}/${this.currencyOut}/${this.tradeType}/${this.chainId}`

cgkol: but you mentioned there are multiple routes for the same key right? THerefore, if we dont go inside and invalidate some of these routes then it wont get fixed?

jsy1218: i see what you saying..

right now the cached routes data structure is that all the routes for the same key will have the same block number that they were cached, [here](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/caching/route/model/cached-routes.ts#L28). but i can see that we should keep each block number, for each route that we stored into this cached routes object, and only invalidate some routes from this cached routes object. 

is that what you are asking? if so, i def agree in principal, but i can elaborate on some implementation difficulties due to how cached routes are implemented in routing-api..

cgkol: yeah... i think thats the right way to go about it, otherwise we lose all routes which is also bad if only one of them is failing right?

src/routers/alpha-router/alpha-router.ts
xrsv: i see that based on [routing api/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L307](https://github.com/Uniswap/routing-api/blob/b1e24e225acf19b6fa3fa171d5a7d513dda25b55/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L307) this caching request is sent every few blocks (3 for mainnet and 15 for others) or if cache entry is missing. is this expected behavior?

src/routers/alpha-router/alpha-router.ts
jsy1218: we discussed offline. current cached routes implementation is sub-optimal, but with time constraint, as well as next gen in mind, we decide to just keep the key level (tokenin/tokenout/tradetype/chainid) to update the cached routes in dynamo. ideally we know we want to update each individual cached routes with its latest block number updated.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1645,6 +1652,102 @@ export class AlphaRouter
       }
     }
 
+    let newSetCachedRoutesPath = false;
+    const shouldEnableCachedRoutesCacheInvalidationFix =
+      Math.random() * 100 <
+      (this.cachedRoutesCacheInvalidationFixRolloutPercentage ?? 0);
```
xrsv, line 1658: eventually this will roll out to 100% of cache entries even with a small percent enabled, if I understand correctly

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1645,6 +1652,102 @@ export class AlphaRouter
...
+              v4GasModel,
+              mixedRouteGasModel,
+              gasPriceWei,
+              v2GasModel,
+              swapConfig,
+              providerConfig
+            ));
+
+          if (swapRouteFromChainAgain) {
+            const routesToCache = CachedRoutes.fromRoutesWithValidQuotes(
```
xrsv, line 1702: seems lines `1702-1746` are duplicated at `1789-1839`.
Can you double check and maybe simplify with a method `cacheAndLog(..., metricName)` or similar?

xrsv (commented):
Approach looks reasonable.
Left a couple of comments

src/routers/alpha-router/alpha-router.ts
jsy1218: yes

xrsv (approved):

src/routers/alpha-router/alpha-router.ts
jsy1218: so yea, originally [filterExpiredCachedRoutes](https://github.com/Uniswap/routing-api/blob/b3025558e807c37e11ef87720b788bdc73d6f4ea/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L483) would filter against the expired cached routes, based on the block number that the cached routes got persisted into the dynamo DB. but with that bug in the expired cached routes, cached routes can remain as obsolete as thousands of blocks just on L1:

Screenshot 2024-10-28 at 1 38 53 PM

jsy1218: now we have a change of plan as discussed last week, we will maintain the mistake in [filterExpiredCachedRoutes](https://github.com/Uniswap/routing-api/blob/b3025558e807c37e11ef87720b788bdc73d6f4ea/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L483), but instead do a cached routes update in the async path in SOR, so that we shouldn't anticipate a online quote endpoint latency increase.

cgkol (approved):
✅

jsy1218:
### Merge activity

* **Oct 28, 5:01 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/smart-order-router/747).

PR #751, opened 2024-10-15
chore: bump sor to 4.7.1 - fix: v4 subgraph provider add ETH base token, and custom fee tier and tick spacings

jsy1218:
Bug fix

Right now v4 is hardcoded with v3 fee tiers and tickspacings. Also static subgraph provider is not able to get native ETH v4 pool.

allow routing-api to pass in more v4 custom fee tiers and tickspacings on a per-need basis. Also static subgraph should get ETH base token to get native v4 pool. Later is in case the v4 subgraph has too many bugs so that we cannot rely on live subgraph to filter by TVL.

- **Other information**:

### DIFF
--- a/src/util/amounts.ts
+++ b/src/util/amounts.ts
@@ -1,10 +1,11 @@
 import { parseUnits } from '@ethersproject/units';
+import { ADDRESS_ZERO } from '@uniswap/router-sdk';
 import {
   ChainId,
   Currency,
   CurrencyAmount as CurrencyAmountRaw,
 } from '@uniswap/sdk-core';
-import { FeeAmount } from '@uniswap/v3-sdk';
+import { FeeAmount, TICK_SPACINGS } from '@uniswap/v3-sdk';
 import JSBI from 'jsbi';
 
 export class CurrencyAmount extends CurrencyAmountRaw<Currency> {}
@@ -72,4 +73,25 @@ export function getApplicableV3FeeAmounts(chainId: ChainId): FeeAmount[] {
   }
 
   return feeAmounts;
+}
+
+export function getApplicableV4FeesTickspacingsHooks(
+  chainId: ChainId
+): Array<[number, number, string]> {
+  const feeAmounts = [
+    FeeAmount.HIGH,
+    FeeAmount.MEDIUM,
+    FeeAmount.LOW,
+    FeeAmount.LOWEST,
+  ];
+
+  if (chainId === ChainId.BASE) {
+    feeAmounts.push(FeeAmount.LOW_200, FeeAmount.LOW_300, FeeAmount.LOW_400);
+  }
+
+  return feeAmounts.map((feeAmount) => [
+    feeAmount as number,
+    TICK_SPACINGS[feeAmount],
+    ADDRESS_ZERO,
+  ]);
 }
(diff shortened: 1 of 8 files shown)


jsy1218:
* **#751** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/751?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/util/amounts.ts
```diff
@@ -73,3 +74,24 @@ export function getApplicableV3FeeAmounts(chainId: ChainId): FeeAmount[] {
...
+    FeeAmount.LOWEST,
+  ];
+
+  if (chainId === ChainId.BASE) {
+    feeAmounts.push(FeeAmount.LOW_200, FeeAmount.LOW_300, FeeAmount.LOW_400);
+  }
+
+  return feeAmounts.map((feeAmount) => [
+    feeAmount as number,
+    TICK_SPACINGS[feeAmount],
```
cgkol, line 94: this is v3 specific, for v4 it can be anything right? how are we planning on accounting for that?

jsy1218: good question. I'm thinking:

1. from LP flow, I see the initial page encourage people to create v4 pools with the same fee-tiers (shown on the UI page) and the same corresponding tick spacings (not shown on the UI page). (This is still beta testing, so I don't wanna post screenshot here. Can DM you.) This means likely new v4 pools through web interface will have the same fee-tiers and tick spacings for majority of the v4 pools.
2. from SOR and routing-api, Im thinking to have SOR level default fee-tier and tick spacings the same as v3. Then at routing-api, I can pass in more fee-tier and tickspacings (and maybe hook address) combos as needed. The way I see this is that we only want to add more fee-tier and tickspacings in adhoc fashion, because otherwise fetching all potential fee-tier and tickspacings for v4 pools is not scalable during routing lambda runtime. There are 10k fee tiers (due to [solidity datatype](https://www.notion.so/Hookless-V4-Routing-8126b60c4b974b55ac5bc3dff4185581?d=d9d9d33ed0694b2ea2af5d3066938f14&pvs=4#ec641262184844cc915fb2f48cabb772)) and 32766 tick spacings ([tickspacing has granularity of two](https://uniswapteam.slack.com/archives/C07AD3507SQ/p1728587844162899), min 1, max 32767), hence 32 millions potential hookless v4 pools already. Routing has no idea which v4 pools out of which those are initialized.

jsy1218:
for tomorrow's web v4 bash, i think this PR is critical to make people be able to swap v4 ETH/USDC pools on sepolia. But we just newly initialized new correct-sqrtPriceX96 ETH/USDC pools with hardcoded hook address 0x0000000000000000000000000000000000000020. right now SOR only searches v4 pool with hook address of `address(0)`. We will have to allow routing-api to pass in the hardcoded hook address on Sepolia in order to be able to swap ETH/USDC pools with hook address 0x0000000000000000000000000000000000000020 tmrw.

On production nets, we shouldn't allow any hook routing yet though.

cc @cgkol @xrsv

src/util/amounts.ts
xrsv: perhaps a 1-pager research on our options/limitations here can help us determine how to best approach this eventually.

xrsv (approved):

jsy1218:
I want to modify the sepolia alpha-router integ test to try native -> USDC with 0x0000000000000000000000000000000000000020 quote and tenderly simulated swap

cgkol (approved):

PR #756, opened 2024-10-30
chore: more granular metrics for routes/quotes fetched

xrsv:
Adds more granular metrics for routes/quotes fetched.
This will give us more visibility on onChain effort, total routes, total quotes fetched per chain.

Only logging number of quotes/latency at the top level 

Logging quotes/routes/latency at top + per chain level.

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/quoters/v2-quoter.ts
+++ b/src/routers/alpha-router/quoters/v2-quoter.ts
@@ -206,11 +206,30 @@ export class V2Quoter extends BaseQuoter<V2CandidatePools, V2Route, Token> {
     );
 
     metric.putMetric(
+      `V2QuotesLoad_Chain_${this.chainId}`,
+      Date.now() - beforeQuotes
+    );
+
+    metric.putMetric(
       'V2QuotesFetched',
       _(routesWithQuotes)
         .map(([, quotes]) => quotes.length)
         .sum(),
       MetricLoggerUnit.Count
+    );
+
+    metric.putMetric(
+      `V2QuotesFetched_Chain_${this.chainId}`,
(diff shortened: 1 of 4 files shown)


src/routers/alpha-router/quoters/v2-quoter.ts
```diff
@@ -213,6 +218,20 @@ export class V2Quoter extends BaseQuoter<V2CandidatePools, V2Route, Token> {
...
     );
 
+    metric.putMetric(
+      `V2QuotesFetched_Chain_${this.chainId}`,
+      _(routesWithQuotes)
+        .map(([, quotes]) => quotes.length)
+        .sum()
+    );
+
+    metric.putMetric(`V2RoutesFetched`, _(routesWithQuotes).sum());
```
jsy1218, line 228: so this logs the routes fetched in case we have a cache routes miss. in case we have a cached routes hit, it will be within [refreshRoutesThenGetQuotes](https://github.com/Uniswap/smart-order-router/pull/756/files#diff-e2eb9d7f54a7e353f88a2a325fb275b4a08e31c49661fa3851bc87429e30d2a6R302).

I'm thinking (1) whether we should log routes and/or quotes fetched in case of cached routes hit (2) whether we should log both cached routes hit and cached routes miss

jsy1218:
good stuff, let's sync on the logging details

src/routers/alpha-router/quoters/v2-quoter.ts
xrsv: makes sense - lets chat

xrsv:
discussed offline - will add logging in a different place

PR #757, opened 2024-10-30
chore: more metrics for routes/quotes fetched

xrsv:
More metrics for routes/quotes fetched

No metrics for number of routes/quotes per chain.

Adding metrics for number of routes/quotes per chain.

- **Other information**:

### DIFF
--- a/src/providers/on-chain-quote-provider.ts
+++ b/src/providers/on-chain-quote-provider.ts
@@ -1245,9 +1245,7 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
         protocol,
         optimisticCachedRoutes
       )}RoutesQuotesLength`,
-      routesQuotes
-        .map((routeWithQuotes) => routeWithQuotes[1].length)
-        .reduce((a, b) => a + b, 0),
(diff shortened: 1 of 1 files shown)


src/providers/on-chain-quote-provider.ts
```diff
@@ -1223,6 +1223,60 @@ export class OnChainQuoteProvider implements IOnChainQuoteProvider {
...
+    metric.putMetric(
+      `${this.metricsPrefix(
+        this.chainId,
+        useMixedRouteQuoter,
+        mixedRouteContainsV4Pool,
+        protocol,
+        optimisticCachedRoutes
+      )}RoutesQuotesLength`,
+      routesQuotes
+        .map((routeWithQuotes) => routeWithQuotes[1].length)
```
jsy1218, line 1249: hmm... total quotes can be simplified to success quotes + failed quotes right? so that we dont have to repeat the logic from `flatMap((routeWithQuotes: RouteWithQuotes<TRoute>) => routeWithQuotes[1])`

jsy1218 (approved):
LGTM, left one nit

PR #770, opened 2024-11-30
fix(subgraph): include top 20 VIRTUAL/* pools in V2 Base subgraph

xrsv:
Allow top 20 VIRTUAL/* pool for V2 on Base subgraph pool filtering.

### DIFF
--- a/src/providers/v2/subgraph-provider.ts
+++ b/src/providers/v2/subgraph-provider.ts
@@ -324,6 +324,7 @@ export class V2SubgraphProvider implements IV2SubgraphProvider {
   }
 
   // TODO: Remove. Temporary fix to allow VIRTUAL/* V2 top pools to be included in the list until the subgraph is re-indexed.
+  // Note: All addresses here should be in lowercase.
   private static readonly VirtualPairBaseV2PoolIds: Set<string> = new Set([
     // Pool ID for VIRTUAL / GAME
     '0xd418dfe7670c21f682e041f34250c114db5d7789',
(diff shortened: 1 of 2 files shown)


src/providers/v2/subgraph-provider.ts
```diff
@@ -317,4 +314,58 @@ export class V2SubgraphProvider implements IV2SubgraphProvider {
...
+    return (
+      this.chainId === ChainId.BASE &&
+      V2SubgraphProvider.VirtualPairBaseV2PoolIds.has(pool.id.toLowerCase())
+    );
+  }
+
+  // TODO: Remove. Temporary fix to allow VIRTUAL/* V2 top pools to be included in the list until the subgraph is re-indexed.
+  private static readonly VirtualPairBaseV2PoolIds: Set<string> = new Set([
+    // Pool ID for VIRTUAL / GAME
+    '0xd418dfe7670c21f682e041f34250c114db5d7789',
```
cgkol, line 329: do we want all these pools, or do we want instead to have just virtual/eth? or will the other one not be found at all anyway so we at least need these?

xrsv: Yes we need those. `VIRTUAL/ETH` already goes through filtering.But it seems that usually best path for `ETH/{VirtualEcosystemToken}` swap goes through VIRTUAL, e.g. `ETH/GAME` swap -> `ETH/VIRTUAL/GAME` so we need to allow top `{VirtualEcosystemToken}/VIRTUAL` pools until subgraph re-indexes which might take a while.

jsy1218 (approved):
We need to ask subgraph team to automate calculating tracked eth TVL

src/providers/v2/subgraph-provider.ts
cgkol: got it, understood

PR #776, opened 2024-12-03
fix: Use syntheticGasCostInTermsOfQuoteToken only if within 30% of gasCostInTermsOfQuoteToken

xrsv:
Addresses bug where we use bad synthetic gas estimate to calculate gas.

### DIFF
--- a/src/routers/alpha-router/gas-models/tick-based-heuristic-gas-model.ts
+++ b/src/routers/alpha-router/gas-models/tick-based-heuristic-gas-model.ts
@@ -213,7 +213,7 @@ export abstract class TickBasedHeuristicGasModelFactory<
               gasCostInTermsOfQuoteToken.asFraction
             ) &&
             gasCostInTermsOfQuoteToken.subtract(syntheticGasCostInTermsOfQuoteToken)
-              .lessThan(gasCostInTermsOfQuoteToken.multiply(0.3))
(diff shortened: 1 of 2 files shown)


src/routers/alpha-router/gas-models/tick-based-heuristic-gas-model.ts
```diff
@@ -205,12 +205,16 @@ export abstract class TickBasedHeuristicGasModelFactory<
         // Note that the syntheticGasCost being lessThan the original quoted value is not always strictly better
         // e.g. the scenario where the amountToken/ETH pool is very illiquid as well and returns an extremely small number
         // however, it is better to have the gasEstimation be almost 0 than almost infinity, as the user will still receive a quote
+        // Only use syntheticGasCostInTermsOfQuoteToken if it's within 30% of the original gasCostInTermsOfQuoteToken as a safeguard.
```
jsy1218, line 208: we should prob log metrics on how often gas cost can be more than 30% different from synthetic gas cost in terms of quote token.

jsy1218 (approved):

src/routers/alpha-router/gas-models/tick-based-heuristic-gas-model.ts
xrsv: discussed offline, adding metric would require breaking if statement in several parts / nested ifs which will add complexity. skipping for now

PR #780, opened 2024-12-06
chore: bump sor to 4.8.7 - feat: tenderly sim status SlippageTooLow

jsy1218:
feature

we don't break down the concrete tenderly simulation onchain revert error

We first determine whether the onchain revert might be due to slippage. there's a possibility it could also be insufficient liquidity. but for now we just return slippage too low

- **Other information**:
looking into adding tests, probably unit tests

### DIFF
--- a/src/providers/tenderly-simulation-provider.ts
+++ b/src/providers/tenderly-simulation-provider.ts
@@ -31,6 +31,7 @@ import {
   initSwapRouteFromExisting,
 } from '../util/gas-factory-helpers';
 
+import { breakDownTenderlySimulationError } from '../util/tenderlySimulationErrorBreakDown';
 import { EthEstimateGasSimulator } from './eth-estimate-gas-provider';
 import { IPortionProvider } from './portion-provider';
 import {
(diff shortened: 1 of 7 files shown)


jsy1218:
* **#780** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/780?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/providers/tenderly-simulation-provider.ts
```diff
@@ -464,6 +464,19 @@ export class TenderlySimulator extends Simulator {
               2
             )}.`
           );
+
+          if ((resp.result[2] as JsonRpcError).error.data) {
```
xrsv, line 468: better to check if `resp.result[2]` exists here

jsy1218: aw it's actually covered in the outer if condition:

```
if (
          !resp ||
          !resp.result ||
          resp.result.length < 3 ||
          (resp.result[2] as JsonRpcError).error
        )
```

src/routers/alpha-router/gas-models/tick-based-heuristic-gas-model.ts
```diff
@@ -208,13 +208,16 @@ export abstract class TickBasedHeuristicGasModelFactory<
         // Only use syntheticGasCostInTermsOfQuoteToken if it's within 30% of the original gasCostInTermsOfQuoteToken as a safeguard.
         if (
           syntheticGasCostInTermsOfQuoteToken !== null &&
-          (gasCostInTermsOfQuoteToken === null || (
-            syntheticGasCostInTermsOfQuoteToken.lessThan(
+          (gasCostInTermsOfQuoteToken === null ||
```
jsy1218, line 211: this file is change after running `npm run fix:prettier`

src/providers/tenderly-simulation-provider.ts
xrsv: that `if` condition can still enter if resp/resp.result/resp.result[2] is not present - so here you want to do something like 
`if (resp.result.length >= 3 && (resp.result[2] as JsonRpcError).error.data) {`

jsy1218: made it more defensive

xrsv (approved):

cgkol (approved):

PR #798, opened 2025-01-15
chore: bump sor to 4.14.1 - add bytes32 non-null terminator logging

jsy1218:
chore

we don't know which token had non-null terminator runtime error

log it so that we know

- **Other information**:

### DIFF
--- a/src/providers/token-provider.ts
+++ b/src/providers/token-provider.ts
@@ -898,12 +898,19 @@ export class TokenProvider implements ITokenProvider {
           continue;
         }
 
+        let symbol;
+
         try {
-          isBytes32
+          symbol = isBytes32
             ? parseBytes32String(symbolResult.result[0]!)
             : symbolResult.result[0]!;
         } catch (error) {
-          if (error instanceof Error && error.message.includes('invalid bytes32 string - no null terminator')) {
+          if (
+            error instanceof Error &&
+            error.message.includes(
+              'invalid bytes32 string - no null terminator'
+            )
+          ) {
             log.error(
               {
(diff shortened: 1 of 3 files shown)


jsy1218:
* **#798** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/798?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/798?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/providers/token-provider.ts
```diff
@@ -898,6 +898,23 @@ export class TokenProvider implements ITokenProvider {
...
+                symbolResult,
+                error,
+              },
+              `invalid bytes32 string - no null terminator`
+            );
+            continue;
+          }
+        }
+
         const symbol = isBytes32
```
xrsv, line 918: should we wrap this call instead?
also do we want to log error and continue? previous behavior is to throw I believe

jsy1218: do you see a syntax sugar, that I don't have to call parseBytes32String twice, but rather try-catch, log the error, and re-throw?

xrsv (approved):

jsy1218:
### Merge activity

* **Jan 15, 4:56 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Uniswap/smart-order-router/798).

PR #806, opened 2025-01-27
chore: bump sor to 4.17.2 - fix: Implement check for sellReverted boolean in transfer detector contract

jsy1218:

- **Other information**:

### DIFF
--- a/src/providers/token-properties-provider.ts
+++ b/src/providers/token-properties-provider.ts
@@ -123,8 +123,8 @@ export class TokenPropertiesProvider implements ITokenPropertiesProvider {
       await Promise.all(
         addressesToFetchFeesOnchain.map((address) => {
           const tokenFee = tokenFeeMap[address];
-          const tokenFeeResultExists: BigNumber | undefined =
-            tokenFee && (tokenFee.buyFeeBps || tokenFee.sellFeeBps);
+          const tokenFeeResultExists: BigNumber | boolean | undefined =
+            tokenFee && ((tokenFee.buyFeeBps || tokenFee.sellFeeBps) || tokenFee.sellReverted);
 
           if (tokenFeeResultExists) {
             // we will leverage the metric to log the token fee result, if it exists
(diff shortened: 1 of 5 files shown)


jsy1218:
* **#806** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/806?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xrsv:
> Shall we add a test? also shall we also change the implementaion on graphQL to reflect the change given that we are reading from there to cache first?

Good call @cgkol - We probably need to make similar changes in the [cached path here](https://github.com/Uniswap/smart-order-router/blob/2e8f72f58bc21f66982601e1385bb26f43aff658/src/providers/token-properties-provider.ts#L81). Plus a [test](https://github.com/Uniswap/smart-order-router/blob/e0f7a7932c6473fc7fce25510ccb337723ca123d/test/unit/providers/token-properties-provider.test.ts) would be nice

jsy1218:
> > Shall we add a test? also shall we also change the implementaion on graphQL to reflect the change given that we are reading from there to cache first?
> 
> Good call @cgkol - We probably need to make similar changes in the [cached path here](https://github.com/Uniswap/smart-order-router/blob/2e8f72f58bc21f66982601e1385bb26f43aff658/src/providers/token-properties-provider.ts#L81). Plus a [test](https://github.com/Uniswap/smart-order-router/blob/e0f7a7932c6473fc7fce25510ccb337723ca123d/test/unit/providers/token-properties-provider.test.ts) would be nice

leaning toward adding a test in routing-api, because the actual contract call will be through GQL. ok to add test in token fetcher class, but that's not being used in routing-api actually.

cgkol (approved):
Fair enough

xrsv (approved):

jsy1218:
it doesn't fix

PR #809, opened 2025-01-30
fix(cache): skip cache if not all protocols are requested

xrsv:

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1426,66 +1426,76 @@ export class AlphaRouter
 
     // Fetch CachedRoutes
     let cachedRoutes: CachedRoutes | undefined;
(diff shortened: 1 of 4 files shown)


cgkol (approved):

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1426,66 +1426,76 @@ export class AlphaRouter
...
+                amount,
+                quoteCurrency,
+                tradeType,
+                [Protocol.V4],
+                await blockNumber,
+                routingConfig.optimisticCachedRoutes
+              ),
+            ]);
+
+          cachedRoutes = new CachedRoutes({
```
jsy1218, line 1468: if condition is still needed to make sure cachedRoutes can be undefined

xrsv: merged main

jsy1218 (approved):

PR #812, opened 2025-01-30
fix: don't explore split routes with native&wrappedNative

xrsv:


test/unit/routers/alpha-router/functions/best-swap-route.test.ts
```diff
@@ -149,6 +193,29 @@ describe('get best swap route', () => {
...
+      USDC_USDT_V4_MEDIUM,
+      UNI_WETH_V4_MEDIUM,
+      UNI_ETH_V4_MEDIUM,
+      DAI_WETH_V4_MEDIUM,
+      DAI_ETH_V4_MEDIUM,
+    ];
+    mockV4PoolProvider = sinon.createStubInstance(V4PoolProvider);
+    mockV4PoolProvider.getPools.resolves(buildMockV4PoolAccessor(mockV4Pools));
+    // Mock getPoolId to return a pool info object with the required properties
+    mockV4PoolProvider.getPoolId.callsFake((pool: any) => ({
```
cgkol, line 214: shall you make the small one and maybe the medium one 2 or something?

test/unit/routers/alpha-router/functions/best-swap-route.test.ts
```diff
@@ -838,4 +981,223 @@ describe('get best swap route', () => {
...
+        amount,
+        percents,
+        routesWithQuotes,
+        TradeType.EXACT_INPUT,
+        ChainId.MAINNET,
+        { ...mockRoutingConfig, distributionPercent: 50 },
+        portionProvider
+      )!;
+
+      // Even though mixing native and wrapped native would give better quotes,
```
cgkol, line 1141: so this test should fail after we remove this constraint right? thats good to keep here and make sure it tests that it is working as expected.

cgkol: wait otherwise, you would expect a mix here? maybe ad a few more comments, we should do 190 but instead we do 180 because its better

test/unit/routers/alpha-router/functions/best-swap-route.test.ts
xrsv: good call - updated

test/unit/routers/alpha-router/functions/best-swap-route.test.ts
xrsv: correct - updated comments

cgkol (approved):

PR #833, opened 2025-03-03
fix: don't use cachedRoutes if intent.caching (1% experiment)

xrsv:
Bug fix

- **What is the current behavior?**
Cached routes are still used when intent.caching

- **What is the new behavior?**
Don't use cachedRoutes if intent.caching
Start with `1%` experiment to validate and measure rpc/latency increase

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -3200,21 +3200,22 @@ export class AlphaRouter
     );
   }
 
-  // Percentage of time we allow using cached routes. We start with 99% to gradually roll out the change.
-  public static readonly CACHED_ROUTES_EXPERIMENT_FLAG_ENABLED_DEFAULT_PERCENTAGE = 0.99;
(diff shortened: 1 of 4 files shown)


src/routers/alpha-router/alpha-router.ts
```diff
@@ -3195,4 +3199,22 @@ export class AlphaRouter
       }
     );
   }
+
+  // Percentage of time we allow using cached routes. We start with 99% to gradually roll out the change.
+  public static readonly CACHED_ROUTES_EXPERIMENT_FLAG_ENABLED_DEFAULT_PERCENTAGE = 0.99;
```
jsy1218, line 3204: nit: it feels more intuitive to have a flag ot enable_no_cached_routes_percent with 0.01.

jsy1218 (approved):

src/routers/alpha-router/alpha-router.ts
xrsv: makes sense - updated

PR #843, opened 2025-03-10
chore: bump sor to 4.20.4 - log cached routes change metrics

jsy1218:
chore

we dont know how often cached routes change

we log metrics to know how often cached routes change

- **Other information**:
we also log the chainId, as well as pair. we dont have the trading amount yet, because theres way too many fanouts.

Tested in routing-api, I can see:
- query string param cachedRoutesRouteIds

- cachedRoutesNotChanged metrics

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -151,10 +151,7 @@ import {
 import { UniversalRouterVersion } from '@uniswap/universal-router-sdk';
 import { DEFAULT_BLOCKS_TO_LIVE } from '../../util/defaultBlocksToLive';
 import { INTENT } from '../../util/intent';
-import {
-  deserializeRouteIds,
-  serializeRouteIds
-} from '../../util/serializeRouteIds';
+import { serializeRouteIds } from '../../util/serializeRouteIds';
 import {
   DEFAULT_ROUTING_CONFIG_BY_CHAIN,
   ETH_GAS_STATION_API_URL,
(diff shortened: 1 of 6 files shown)


jsy1218:
* **#843** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/843?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/843?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2020,9 +2026,40 @@ export class AlphaRouter
...
     tradeType: TradeType,
     metricsPrefix: string,
-    routesToCache?: CachedRoutes
+    routesToCache?: CachedRoutes,
+    cachedRouteRouteId?: number
   ): Promise<void> {
     if (routesToCache) {
+      const cachedRoutesChanged =
+        cachedRouteRouteId !== undefined &&
+        cachedRouteRouteId !== routesToCache.routes[0]?.routeId;
```
xrsv, line 2035: so here you check that the routeId of the first route changed? what if it's in a different position?

jsy1218: modified to serialize all the routeids across cached routes

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2020,9 +2027,42 @@ export class AlphaRouter
...
-    routesToCache?: CachedRoutes
+    routesToCache?: CachedRoutes,
+    cachedRoutesRouteIds?: string
   ): Promise<void> {
     if (routesToCache) {
+      const cachedRoutesChanged =
+        cachedRoutesRouteIds !== undefined &&
+        // it's possible that top cached routes may be split routes,
+        // so that we always serialize all the top 8 retrieved cached routes vs the top routes.
+        !cachedRoutesRouteIds.startsWith(serializeRouteIds(routesToCache.routes.map((r) => r.routeId)));
```
xrsv, line 2038: i assume this should work as the order in which we fetch the top N routes (by tvl), should match the order of our splits? If you want to be more flexible, you can check that each routesToCache (e.g. 3 entries) are in the first positions (e.g first 3)  of cachedRoutesRouteIds which covers mixed order as well.

jsy1218: right, so string `startsWith` does the trick of that. It's probably a bit less readable than checking `routeToCache` array against `cachedRouteRouteIds` array though.

PR #860, opened 2025-04-25
feat: skip cachedRoutes for large swaps

xrsv:
fix/feature

always use cachedRoutes

we skip cachedRoutes when input amount in USD > 10K

added unit tests + e2e tests

image


src/util/amounts.ts
```diff
@@ -95,3 +95,6 @@ export function getApplicableV4FeesTickspacingsHooks(
     ADDRESS_ZERO,
   ]);
 }
+
+// Used in order to skip cached routes if input amount is larger than this amount.
+export const LARGE_SWAP_USD_THRESHOLD = 10000;
```
jsy1218, line 100: i recommend passing this threshold from routing-api `AlphaRouterConfig`, so that in future, we can lower the amount easily in routing-api. Also, I wonder if we should start with a higher amount?

jsy1218 (approved):

src/providers/token-price-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+export class TokenPriceProvider implements ITokenPricesProvider {
+  private CACHE_KEY = (chainId: ChainId, address: string) =>
+    `token-price-${chainId}-${address}`;
+
+  constructor(
+    private chainId: ChainId,
+    private tokenPricesCache: ICache<TokenPriceResult>,
+    private tokenPriceFetcher: ITokenPriceFetcher,
+    private positiveCacheEntryTTL = POSITIVE_TOKEN_PRICE_CACHE_ENTRY_TTL,
+    private negativeCacheEntryTTL = NEGATIVE_TOKEN_PRICE_CACHE_ENTRY_TTL
```
cgkol, line 45: negative meaning it does not exist?

src/routers/alpha-router/alpha-router.ts
```diff
@@ -3269,12 +3285,77 @@ export class AlphaRouter
     );
   }
 
-  // We want to skip cached routes access whenever "intent === INTENT.CACHING" or "hooksOption !== HooksOption.NO_HOOKS"
+  // Check if the given amount is considered a large swap (input amount >= LARGE_SWAP_USD_THRESHOLD)
+  private async isLargeSwap(amount: CurrencyAmount): Promise<boolean> {
```
cgkol, line 3289: async does not hit this path right?

cgkol (approved):
Couple of comments + siyu's, looks good overall

src/routers/alpha-router/alpha-router.ts
xrsv: it is hit by async as well.
my thinking is that we can eventually use this to do some extra work on async path (e.g. search deeper)
I can separate to not hit it on async but logic will get a bit complex - results are cached fro 5mins anyway so no real latency impact.

src/providers/token-price-provider.ts
xrsv: means we couldn't get a price from gql, so we add an undefined value in cache with this ttl (it's the same ttl as positive for now)

src/util/amounts.ts
xrsv: updated

src/routers/alpha-router/alpha-router.ts
cgkol: but on async we want to run these anyway right?

xrsv: on async (INTENT.CACHING) we always skip cache anyway - so it doesn't really matter if it's a small/large swap

jsy1218:
* **#860** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/860?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/860?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #863, opened 2025-04-30
chore: bump sor to 4.21.4 - fix: tenderly simulation during caching lambda

jsy1218:

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1928,17 +1928,21 @@ export class AlphaRouter
       );
 
       if (routingConfig.intent === INTENT.CACHING) {
-        await this.performCachingIntention(
-          currencyIn,
-          currencyOut,
-          tradeType,
-          amount,
-          protocols,
-          await blockNumber,
-          routingConfig,
-          swapRouteWithSimulation.simulationStatus!,
-          swapRouteWithSimulation.route
-        );
+        if (swapRouteWithSimulation.simulationStatus) {
+          await this.performCachingIntention(
+            currencyIn,
+            currencyOut,
+            tradeType,
+            amount,
+            protocols,
+            await blockNumber,
+            routingConfig,
+            swapRouteWithSimulation.simulationStatus!,
+            swapRouteWithSimulation.route
+          );
+        } else {
+          log.info('Simulation status is null');
+        }
       }
 
       return swapRouteWithSimulation;
@@ -1974,16 +1978,7 @@ export class AlphaRouter
     // due to fire and forget nature, we already take note that we should set new cached routes during the new path
     metric.putMetric(`SetCachedRoute_NewPath`, 1, MetricLoggerUnit.Count);
 
-    if (simulationStatus) {
-      log.info(
-        {
-          simulationStatus: simulationStatus,
-        },
-        'Simulation status exists'
-      );
-    } else {
-      log.info('Simulation status is null');
-    }
+    log.info(simulationStatus, 'Simulation status');
 
     if (this.shouldCacheRoute(simulationStatus)) {
       const routesToCache = CachedRoutes.fromRoutesWithValidQuotes(
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -1937,10 +1937,6 @@ export class AlphaRouter
         MetricLoggerUnit.Milliseconds
       );
 
-      // we have to write cached routes right before checking swapRouteRaw is null or not
-      // because getCachedRoutes in routing-api do not use the blocks-to-live to filter out the expired routes at all
-      // there's a possibility the cachedRoutes is always populated, but swapRouteFromCache is always null, because we don't update cachedRoutes in this case at all,
-      // as long as it's within 24 hours sliding window TTL
       if (routingConfig.intent === INTENT.CACHING) {
         // due to fire and forget nature, we already take note that we should set new cached routes during the new path
         metric.putMetric(
--- a/src/routers/alpha-router/alpha-router.ts
+++ b/src/routers/alpha-router/alpha-router.ts
@@ -7,7 +7,7 @@ import {
   Currency,
   Fraction,
   Token,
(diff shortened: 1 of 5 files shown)


jsy1218:
* **#863** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2066,6 +1959,60 @@ export class AlphaRouter
...
+
+      const shouldEnableCachedRoutesCacheInvalidationFix =
+        Math.random() * 100 <
+        (this.cachedRoutesCacheInvalidationFixRolloutPercentage ?? 0);
+
+      // we have to write cached routes right before checking swapRouteRaw is null or not
+      // because getCachedRoutes in routing-api do not use the blocks-to-live to filter out the expired routes at all
+      // there's a possibility the cachedRoutes is always populated, but swapRouteFromCache is always null, because we don't update cachedRoutes in this case at all,
+      // as long as it's within 24 hours sliding window TTL
+      if (shouldEnableCachedRoutesCacheInvalidationFix) {
```
cgkol, line 1971: to simplify the logic, shall we remove this if statement since its always on anyway?

cgkol (commented):
i think the PR is valid other than that

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2066,6 +1959,60 @@ export class AlphaRouter
...
+        // theoretically, when routingConfig.intent === INTENT.CACHING, optimisticCachedRoutes should be false
+        // so that we can always pass in cachedRoutes?.notExpired(await blockNumber, !routingConfig.optimisticCachedRoutes)
+        // but just to be safe, we just hardcode true when checking the cached routes expiry for write update
+        // we decide to not check cached routes expiry in the read path anyway
+        if (!cachedRoutes?.notExpired(await blockNumber, true)) {
```
xrsv, line 1976: is this check needed? we always want to update cached routes with new routes, correct?
also since we only update cache in async, this should now be undefined. Let me know if I miss something.

Other than that it looks ok

src/routers/alpha-router/alpha-router.ts
yucanliu: the if statement is removed

src/routers/alpha-router/alpha-router.ts
yucanliu: the check is removed

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2043,6 +1936,42 @@ export class AlphaRouter
...
+        metric.putMetric(
+          `SetCachedRoute_NewPath`,
+          1,
+          MetricLoggerUnit.Count
+        );
+
+        // only when tenderly sim failed for unknown reason, then we do not want to cache the routes
+        // otherwise, if it fails for known reasons 1) insufficient balance 2) system down 3) slippage 4) transfer from failed
+        // we still want to cache the routes. since those are surfaced to FEs through trading-api, still worth caching the routes
+        if (swapRouteWithSimulation.simulationStatus !== SimulationStatus.Failed) {
```
jsy1218, line 1951: thinking whether we need to prevent caching routes when tenderly status is 1) system down 2) slippage 3) transfer from failed.

- system down, we might want to cache it, because tenderly server might just be down temporarily. We don't want to retry against tenderly though.
- slippage, we might not want to cache it, because we know swap is bound to fail
- transfer from failed, we might not want to cache it, because we know swap is bound to fail

test/unit/routers/alpha-router/alpha-router.test.ts
```diff
@@ -1498,7 +1498,7 @@ describe('alpha router', () => {
         expect(swap).toBeDefined();
```
jsy1218, line 1499: we need additional unit test in this file, with the set up to have tenderly status returning SUCCESS, so that `inMemoryRouteCachingProvider.internalSetCacheRouteCalls` is invoked at least once

src/routers/alpha-router/alpha-router.ts
yucanliu: makes sense.

Added logic to NOT cache:
1. failed
2. slippage
3. transfer from failed

When "System down", cache will still be logged.

test/unit/routers/alpha-router/alpha-router.test.ts
yucanliu: test added!

src/routers/alpha-router/alpha-router.ts
```diff
@@ -2043,12 +1926,105 @@ export class AlphaRouter
...
+      if (routingConfig.intent === INTENT.CACHING) {
+        await this.performCachingIntention(
+          currencyIn,
+          currencyOut,
+          tradeType,
+          amount,
+          protocols,
+          await blockNumber,
+          routingConfig,
+          swapRouteWithSimulation.simulationStatus!,
```
xrsv, line 1939: are we sure `swapRouteWithSimulation.simulationStatus!` is always defined?
probably worth passing it as `simulationStatus?: SimulationStatus` and skip cache if undefined?

xrsv (commented):
looks good - left one comment

src/routers/alpha-router/alpha-router.ts
jsy1218: I think it's always defined now, as part of trading-api's shared swap, where trading-api always trigger tenderly sim.

yucanliu: make sense - I updated the logic to check simulationStatus is not null before calling `performCachingIntention` and log a line if it is. Please lmk if this looks good to you

xrsv (approved):
looks good

PR #879, opened 2025-05-22
fix: ROUTE-489 Refresh pools in routes' metadata

yucanliu:
- **What kind of change does this PR introduce?** (Bug fix)
This PR refreshes the pools in the route object from swapRouteFromCache.
The offline lambda is never in the cache mode and the refresh helps to prevent stale data.

https://linear.app/uniswap/issue/ROUTE-489/investigate-cachedroutes-pool-info-refresh-rate

The pools in the route object will be refreshed

- **Other information**:


jsy1218 (approved):

src/util/mutable.ts
```diff
@@ -0,0 +1,14 @@
...
+ * This is useful for mutating objects in place.
+ *
+ * Warning: This should be used with caution since it can lead to unexpected behavior.
+ *
+ * @example
+ * const obj: Mutable<{ readonly prop: string }> = { prop: 'value' };
+ * obj.prop = 'newValue';
+ */
+
+export type Mutable<T> = {
```
cgkol, line 12: dont love this tbh, but if this works fine lets do it.

also @xrsv shall we add it as a ticket for uniroute as well?

cgkol (approved):

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1696,6 +1711,16 @@ export class AlphaRouter
     let swapRouteRaw: BestSwapRoute | null;
     let hitsCachedRoute = false;
     if (cacheMode === CacheMode.Livemode && swapRouteFromCache) {
+      // offline lambda is never in cache mode
+      // refresh pools to avoid stale data
+      await this.refreshPools(
```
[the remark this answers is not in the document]
jsy1218: i think we only want to update pool metadata for the best selected route, in case the best selected route is a cached route. This saves quite a bit of RPC calls.

src/routers/alpha-router/alpha-router.ts
```diff
@@ -1951,6 +1976,205 @@ export class AlphaRouter
...
+    v2PoolProvider: IV2PoolProvider
+  ) {
+    const refreshedPairs: V2Pool[] = [];
+    for (const pair of route.pairs) {
+      const v2Pools = await v2PoolProvider.getPools(
+        [[pair.token0, pair.token1]],
+        config
+      );
+      const refreshed = v2Pools.getPool(pair.token0, pair.token1);
+      if (refreshed) refreshedPairs.push(refreshed);
```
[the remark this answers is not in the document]
jsy1218: i see, yes, we need to make sure we still give back the pool. we prob dont want to throw exception here, because that will make the quote response 5xx. but between incorrect pools[] array (by missing pool object in the array), and stale pools[] array info, I think we should prefer the later

jsy1218, line 2051: if (refreshed) { 
  refreshedPairs.push(refreshed);
} else {
  refreshedPairs.push(pair);
}

src/util/mutable.ts
yucanliu: I don't like this hacky solution neither. 

Originally, I chatted with Siyu and voted for the less code option. But since both you and Vasili had objections, I converted to the reconstruction method.

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
yucanliu: updated - now we will log an error and pushback the original pool.

src/routers/alpha-router/alpha-router.ts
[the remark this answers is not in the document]
yucanliu: add metric around the `refreshPools` method

src/routers/alpha-router/alpha-router.ts
yucanliu: updated :)

src/routers/router.ts
```diff
@@ -255,3 +257,51 @@ export abstract class ISwapToRatio<RoutingConfig, SwapAndAddConfig> {
...
+export function cloneMixedRouteWithNewPools(
+  originalRoute: MixedRoute,
+  newPools: TPool[]
+): MixedRoute {
+  // Reuse the original input and output currencies
+  const input = originalRoute.input;
+  const output = originalRoute.output;
+
+  // Construct a new Route instance with the new pairs
+  return new MixedRoute(newPools, input, output /* retainFakePools = true*/);
```
yucanliu, line 306: @jsy1218 @xrsv do you know if I should set `retainFakePools` to `true` here?

xrsv: based on [this](https://github.com/Uniswap/sdks/blob/730f0cd3e9ab75f561e97a0f7b77d7f42197d813/sdks/router-sdk/src/entities/mixedRoute/route.ts#L30), I believe we want to explicitly set it to true - make sure we don't filter out anything on copy. I'll let @jsy1218  confirm

src/util/mutable.ts
xrsv: @cgkol pool refresh is already in place in uniroute

xrsv (approved):
👍

jsy1218 (approved):

src/routers/alpha-router/quoters/v3-quoter.ts
```diff
@@ -99,7 +99,8 @@ export class V3Quoter extends BaseQuoter<V3CandidatePools, V3Route, Token> {
         // ROUTE-495 - bypass v3 pool FOT check for $ELMO token
         if (
           tokenValidation == TokenValidationResult.FOT &&
-          token.wrapped.address.toLowerCase() === '0x335f4e66b9b61cee5ceade4e727fcec20156b2f0' &&
+          token.wrapped.address.toLowerCase() ===
```
yucanliu, line 102: format change only

jsy1218 (approved):

PR #884, opened 2025-05-30
chore: bump sor to 4.21.18 - fix: invariant token because of revisited v4 native pool eth/weth

jsy1218:
Bug fix

We have invariant token error, because in v4, we can have circular route because of native & wrapped. e.g. ETH/OP -> ETH/OP-> ETH/opXEN

We make sure in case of v4, we track both native and wrapped as the same token being visited.

- **Other information**:
in case of mixed, we need to make sure we don't track both native and wrapped as the same token being visited.

### DIFF
--- a/src/routers/alpha-router/functions/compute-all-routes.ts
+++ b/src/routers/alpha-router/functions/compute-all-routes.ts
@@ -1,4 +1,4 @@
-import { ADDRESS_ZERO, TPool } from '@uniswap/router-sdk';
+import { ADDRESS_ZERO, Protocol, TPool } from '@uniswap/router-sdk';
 import { ChainId, Currency, Token } from '@uniswap/sdk-core';
 import { Pair } from '@uniswap/v2-sdk';
 import { Pool as V3Pool } from '@uniswap/v3-sdk';
@@ -7,7 +7,7 @@ import { Pool as V4Pool } from '@uniswap/v4-sdk';
 import {
   getAddressLowerCase,
   nativeOnChain,
-  V4_ETH_WETH_FAKE_POOL,
+  V4_ETH_WETH_FAKE_POOL
 } from '../../../util';
 import { HooksOptions } from '../../../util/hooksOptions';
 import { log } from '../../../util/log';
@@ -17,7 +17,7 @@ import {
   SupportedRoutes,
   V2Route,
   V3Route,
-  V4Route,
+  V4Route
 } from '../../router';
 
 export function computeAllV4Routes(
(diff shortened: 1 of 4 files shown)


jsy1218:
* **#884** <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/smart-order-router/884?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/routers/alpha-router/functions/compute-all-routes.ts
```diff
@@ -230,11 +235,22 @@ export function computeAllRoutes<
         ? curPool.token1
         : curPool.token0;
 
-      if (tokensVisited.has(getAddressLowerCase(currentTokenOut))) {
+      // In case of protocol mixed, we need to ensure we distinguish between native and wrapped native
+      // because we inject ETH-WETH fake pool to connect mixed routes
+      // Otherwise, in v2,v3,v4, we can just use the wrapped address
+      // we have existing unit test 'handles ETH/WETH wrapping in mixed routes' coverage
+      // in case someone changes the logic to remove MIXED special case
+      const currentTokenVisited = protocol === Protocol.MIXED ? getAddressLowerCase(currentTokenOut) : currentTokenOut.wrapped.address;
```
xrsv, line 243: don't we inject it on V4 routes as well?

xrsv: got it - where is this check done (that it's MIXED) when we inject the fake pool?
I see below which should add on V4 only as well?

```
  const containsV4NativePools =
    parts.filter(
      (pool) =>
        pool instanceof V4Pool &&
        pool.v4InvolvesToken(nativeOnChain(currencyIn.chainId))
    ).length > 0;
  const amendedPools =
    containsV4NativePools && shouldEnableMixedRouteEthWeth
      ? parts.concat(V4_ETH_WETH_FAKE_POOL[currencyIn.chainId as ChainId])
      : parts;
```

cgkol (approved):
lets do it!

src/routers/alpha-router/functions/compute-all-routes.ts
jsy1218: correct. because when there's no v4 pool, there's no way v2 or v3 pool can be native pool.

PR #885, opened 2025-06-12
chore: bump sor to 4.21.19 - fix: add eth/weth fake pool to v4 routes siyujiang/eth-weth-fake-pool-v4

jsy1218:
Bug fix

We still have invariant token error when it's v4 only route

We need to create fake ETH/WETH pool, so that we dont have circular route when it's v4 only.

- **Other information**:

### DIFF
--- a/src/routers/alpha-router/functions/compute-all-routes.ts
+++ b/src/routers/alpha-router/functions/compute-all-routes.ts
@@ -7,7 +7,7 @@ import { Pool as V4Pool } from '@uniswap/v4-sdk';
 import {
   getAddressLowerCase,
   nativeOnChain,
-  V4_ETH_WETH_FAKE_POOL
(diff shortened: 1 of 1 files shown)


jsy1218:
we can't have fake eth/weth pool in v4 only, because v4 onchain [quoter](https://github.com/Uniswap/v4-periphery/blob/main/src/lens/V4Quoter.sol) doesn't have a encoded way to know its a fake pool.  in case of mixed [quoter](https://github.com/Uniswap/mixed-quoter/blob/main/src/MixedRouteQuoterV2.sol), we encode protocol version as [0](https://github.com/Uniswap/mixed-quoter/blob/main/src/MixedRouteQuoterV2.sol#L196) to indicate eth/weth fake pool being passed in.

PR #901, opened 2025-07-12
chore: Increased pageSize for Base (all protocols)

xrsv:
V2 cron subgraph started timing out as well due to lots of pools.
Increasing pagesize similarly to v3/v4.


yucanliu (approved):
LGTM - Thanks for updating this! You might need to merge and release this before updating the Routing-api

PR #909, opened 2025-08-08
chore: bump sor to 4.22.13 - fix: support new gql endpoint for zora siyujiang/new-gql-endpointt

jsy1218:
Bug fix

zora has many creator pools and post pools. current subgraph cannot sync to tip. we need to migrate to new subgraph endpoint.

migrate to new subgraph endpoint. new endpoint needs bearer token, meanwhile old one doesn't

- **Other information**:

### DIFF
--- a/src/providers/subgraph-provider.ts
+++ b/src/providers/subgraph-provider.ts
@@ -66,13 +66,23 @@ export abstract class SubgraphProvider<
     private trackedEthThreshold = 0.01,
     // @ts-expect-error - kept for backward compatibility
     private untrackedUsdThreshold = Number.MAX_VALUE,
-    private subgraphUrl?: string
+    private subgraphUrl?: string,
+    private useNewEndpoint = false,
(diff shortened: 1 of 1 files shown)


src/providers/subgraph-provider.ts
```diff
@@ -66,13 +66,23 @@ export abstract class SubgraphProvider<
     private trackedEthThreshold = 0.01,
     // @ts-expect-error - kept for backward compatibility
     private untrackedUsdThreshold = Number.MAX_VALUE,
-    private subgraphUrl?: string
+    private subgraphUrl?: string,
+    private useNewEndpoint = false,
```
xrsv, line 70: `useNewEndpoint` not really needed. we should pass the endpoint
then if bearerToken is defined, we use it
e.g. 
```
if (this.bearerToken) {...}
```

xrsv (approved):

PR #928, opened 2025-10-22
chore(ci): Add trusted publishing

dgilmanuni:
Releases are authenticated via trusted publishing

- **Other information**:
https://docs.npmjs.com/trusted-publishers/#how-trusted-publishing-works


yucanliu (commented):
Thanks, David. I believe this is the right thing to do. However, @jsy1218 can you please confirm as I don't know if it will break any existing release flow

jsy1218 (approved):

PR #937, opened 2025-11-05
fix: pass down currencyIn and currencyOut to cached routes

jsy1218:
Bug fix

in alpha-router, [getCachedRoutes](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/alpha-router.ts#L1559-L1587) for native, we get wrapped native by wrapping both amount currency and quote currency, and native by passing down amount and quote currency. This effectively becomes two async caching lambda request. To note, both async caching lambda will send down WETH as tokenInAddress. This is because in [_getCachedRoute](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L122), we get the token address, in case V4 is included, inside [deriveCurrencyAddress](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/router-entities/route-caching/model/pair-trade-type-chain-id.ts#L44). Now looking back to alpha-router [getCachedRoute](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/alpha-router.ts#L1575) for only getting native, we pass down [V4] protocol, so V4 is included.

Since there are two async caching lambda call, we'd be on the safe side to pass both lambda invocation as tokenInAddress of native ("ETH"). However, we'd be extremely cautious about changing the logic in routing-api, [determining the currencyIn and currencyOut](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L117), because we change the main cached routes partition key lookup logic to lookup by native instead of wrapped, we will hit cache miss in [_getCachedRoute](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L207), and otherwise not able to send an async lambda invocation. Therefore the best option is for SOR to send the original currencyIn and currencyOut to routing-api, and routing-api basically pass in the original tokenInAddress and tokenOutAddress as they would get from the quote request.

- **Other information**:
there's more. [v3 quoter](https://github.com/Uniswap/view-quoter-v3) does not detect the insffucient liuqidity issue, because it's using cache. meanwhile [v4-quoter](https://github.com/Uniswap/v4-periphery/blob/main/src/lens/V4Quoter.sol) actually simulate swap through pool manager, so in case of insufficient liquidity it will revert. In case of 100 ETH -> ISLAND, no v3 or v4 pool has that amount of liquidity. So in the async lambda path, v3 quoter somehow can still give back the quote, meanwhile v4 quoter reverted due to insffucient liquidity. So in this case, async routing lambda still inserts the same v3 route into the cached routes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@uniswap/smart-order-router",
-  "version": "4.22.27",
+  "version": "4.22.28",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@uniswap/smart-order-router",
-      "version": "4.22.27",
+      "version": "4.22.28",
       "license": "GPL",
       "dependencies": {
         "@eth-optimism/sdk": "^3.2.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@uniswap/smart-order-router",
-  "version": "4.22.27",
+  "version": "4.22.28",
   "description": "Uniswap Smart Order Router",
   "main": "build/main/index.js",
   "typings": "build/main/index.d.ts",


src/providers/caching/route/route-caching-provider.ts
```diff
@@ -193,6 +199,8 @@ export abstract class IRouteCachingProvider {
     tradeType: TradeType,
     protocols: Protocol[],
     currentBlockNumber: number,
+    currencyIn: Currency,
```
cgkol: this PR currently doesnt change nay logic it just adds fields right?

cgkol: oh i guess the llogic chane would be on routing-api i see

jsy1218: thats right

PR #938, opened 2025-11-06
fix: monad tenderly not supported

jsy1218:
Bug fix

Monad is not supported on tenderly yet

Tenderly trying to get early access to Monad, but no guarantee availability upon monad GA

- **Other information**:


yucanliu (approved):
LGTM, thanks!

